PR #51, opened 2019-02-18
Feature/webpack-frontend

edwinmp:


akmiller01 (approved):
Looks good @edwinmp. We're you able to change urls.py and get routing to work?

edwinmp:
Since we'll be doing the main routing through `react-router`, the only way they can work together is if `react-router` paths are duplicated in `urls.py` with the same destination; `views.index`. This should push them to react-router for when the page is refreshed. We could use some regex to redirect all routes this way in one line, but I think that may affect the API routes... not sure.

PR #52, opened 2019-02-21
feature/login

edwinmp:
- [x] Local Login
- [ ] Google Login


akmiller01 (dismissed):
Looks good. Would it be worth looking into adding tests with CI @edwinmp ? Maybe I can help with that next week.

edwinmp:
@akmiller01 have set up travis to run unit tests, though only successfully for the JS tests. Tried to run `python` tests today, but they only run successfully within the `docker` container (only locally). It's failing on travis. Will continue looking into it tomorrow. Also going to start on the end-to-end tests tomorrow.

akmiller01 (dismissed):

akmiller01 (approved):
Looks good! We may want to consider removing some of the hanging dependencies if we don't end up going the social route.

edwinmp:
True ... will put a deadline on it.

PR #54, opened 2019-02-28
Feature/models

Napho:
Only consider changes in database_setup, core/models.py and core/models_templates.py. All other changes can be dropped


akmiller01:
Closing due to .gitignore not working as expected. Using branch feature/models-clean instead https://github.com/devinit/ddw-analyst-ui/tree/feature/models-clean

edwinmp:
I think `.gitignore` was working fine... it just can't ignore files that are already pushed. You'd have to delete them first, then push.

akmiller01:
When you add files @Napho , are you doing like `git add -A`? I do that all the time when I'm working solo, but on group projects it's good to try and add things individually.

PR #55, opened 2019-02-28
Clean models branch

Napho:
Added models from inspectdb for mirror databases and a few database first tables. 
Made modification to database creation scripts to accommodate changes in schemas
Separated database connection settings to allow for running migration and inspectdb concurrently


Napho:
Something for us to think about though
During DB setup, i separated the scripts into init.sh and loader.sh
loader.sh can only be run after python migrations are done because loader.sh creates tables that reference python migrated tables

But using docker service there is not way of separating these concerns

akmiller01:
Hey @edwinmp , might be worth looking into how Husky is working still. I couldn't get it to push without `--no-verify` until I changed all references in `package.json` from `python` to explicitly `python3` and install all of the requirements in my system pip. Even with the virtualenv activated, Husky was ignoring it and using my system python (which defaults to 2.7).

core/models_template.py
```diff
@@ -0,0 +1,314 @@
...
+    year = models.SmallIntegerField()
+    value = models.DecimalField(max_digits=65535, decimal_places=65535, blank=True, null=True)
+    flags = models.TextField(blank=True, null=True)
+
+    class Meta:
+        managed = False
+        db_table = 'dac5_current'
+
+
+class Operation(models.Model):
```
akmiller01: @Napho can we move these into `models.py` and make them managed by Django's ORM directly? Even after following the DB setup steps, we cannot add them to admin.py without the following error message: 
image

core/models_template.py
```diff
@@ -0,0 +1,314 @@
...
+    is_draft = models.BooleanField()
+    created_on = models.DateTimeField()
+    updated_on = models.DateTimeField(blank=True, null=True)
+
+    class Meta:
+        managed = False
+        db_table = 'operation'
+
+
+class OperationSteps(models.Model):
```
akmiller01: I think all of these non-data models belong in `models.py`

akmiller01 (changes requested):

database_setup/test_user_accounts.sql
```diff
@@ -0,0 +1,15 @@
...
+\.
+
+
+--
+-- TOC entry 2527 (class 0 OID 0)
+-- Dependencies: 228
+-- Name: auth_user_id_seq; Type: SEQUENCE SET; Schema: user_mgnt; Owner: postgres
+--
+
+SELECT pg_catalog.setval('auth_user_id_seq', 4, true);
```
edwinmp, line 15: I'm going to bother you about small things. Try to end every file with a new line... it's conventional. Github complains about it as well... notice the tiny red thingie at the end of this line.

database_setup/README.md
```diff
@@ -0,0 +1,19 @@
+## Development Database Setup
+
+To create a test development DB,
```
edwinmp, line 3: Had tried to move the database setup info into the main README, but it's fine if it stays here... just have a link to it on the main README under the same title.

ddw_analyst_ui/settings.py
```diff
@@ -99,7 +99,7 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql',
-        'NAME': 'postgres',
+        'NAME': 'analyst_ui',
```
edwinmp, line 102: There's two `settings.py` files. This is the main one you can use locally, then there's one for `Docker` (. Please make this update to the `Docker` file as well `docker_settings.py`.

package.json
```diff
@@ -6,15 +6,15 @@
   "scripts": {
     "dev": "grunt",
     "build": "grunt release",
-    "test": "jest -u --coverage && python manage.py test",
-    "test:docker": "jest -u && docker-compose run web python manage.py test",
-    "collectstatic": "python manage.py collectstatic --no-input",
+    "test": "jest -u --coverage && python3 manage.py test",
```
edwinmp: I'd rather this stayed as `python` ... up to you to point the `python` command to run your preferred version of python. It's future proof. Would rather not have to change this to python4 whenever that version gets out.

https://askubuntu.com/questions/320996/how-to-make-python-program-command-execute-python-3

edwinmp (changes requested):

core/models_template.py
Napho: Cool, for ease of use i wiill do that, though i think the query is failing because its targeting the public schema

ddw_analyst_ui/settings.py
Napho: Second settings was added only for migration/inspectdb activities which wasn't a managed process (docker's scope), wont be necessary though since I'm moving the models to be managed by django

core/models_template.py
Napho: I'll just move everything to be managed by django and reduce on the complexity of our process

Napho:
Closing request and changing models work flow on a new branch

akmiller01 (approved):

package.json
akmiller01: Darn, couldn't merge this in until you also approve @edwinmp. As @Napho said in Slack, it's not much of a concern because python2 is incompatible with python3 which will be incompatible with python4. In fact, this incompatibility is leading to my local install complaining when you try and run Python3 Django with the system python3.

edwinmp (approved):

PR #56, opened 2019-02-28
Suggested changes to make metadata tables managed

akmiller01:
Hey @Napho , here's what the Python would look like to make the tables managed as I commented on https://github.com/devinit/ddw-analyst-ui/pull/55 This will still require changes to `database_schema.sql` to remove the pieces that create those in the `table_meta` schema and will probably instead make them in `public`. Thoughts?

PR #57, opened 2019-03-01
Feature/main-layout

edwinmp:
Opted for a sidebar as our main navigation component. It's retractable so no worries about space.


akmiller01 (dismissed):
Looks good! 
image

edwinmp:
Ah, thanks for reminding me of this feature on small devices. That background should not be black ... and the logout options isn't available in this view. Let me add those.

akmiller01 (dismissed):
Looking food! I still had to edit my `settings.py` to get it to migrate, but hopefully the setup will be fixed in #56 

image

akmiller01 (approved):
Yep! Looking good!

PR #58, opened 2019-03-04
Feature/audit trail

akmiller01:
Branched off of #55 since I wasn't able to merge that in yet.

This creates a post-save and pre-delete signal which catches all models in the core app and creates a log. It also adds `user` as a field to every model in core, since it's not possible to catch the current user in a thread-save way in the signals. We'll need to make sure the REST API assigns the current `request.user` to the `user` field of each model on save (without overriding the default `save` function otherwise we might skip the `post_save` signal).

Included one test which just sees whether an audit is created when a new Tag is created.


edwinmp:
Have merged #55 and rebased `develop` into this branch.

This seemed like one of those features someone out there has needed and built, then shared. So please have a look at the options [here](https://djangopackages.org/grids/g/model-audit/). [django-revision](https://django-reversion.readthedocs.io/en/stable/) seems to cover what we need for this use case, and then some. [django-reversion](https://django-reversion.readthedocs.io/en/stable/) is interesting as well, in a different way though.

akmiller01:
I don't think `django-revision` fits our use-case @edwinmp. Maybe we need to have a discussion about this feature, because I thought it made sense to have all of the audit logs under one model so they can all be viewed under one admin page. And it would be storing a lot of extra data to actually keep all models version-controlled. Plus, I generally prefer to code my own features over third-party plugins.

edwinmp:
Oh, definitely not `django-reversion` in regards to this feature. It's interesting, though. I seem to have got the link wrong ... meant [django-simple-history](https://github.com/treyhunner/django-simple-history). Sorry.
As for preferring to code your own feature, that's alright as well, but it's not the practical path with time sensitive projects. Personal ones, sure ... though without open source, projects that take months to build would take years. So in my experience, first instinct is to not try to reinvent the wheel unless what exists just doesn't cut it. Choice is yours though.

akmiller01:
Hm, `django-simple-history` seems nice but it's still not centralized. Each history still seems to exist under the respective model (or even the respective instance) rather than in a centralized place.

edwinmp:
@akmiller01 Any idea why we can't seem to see dateTime fields on the Django admin?

edwinmp (dismissed):

akmiller01:
Oh good point @edwinmp. They're auto-add fields, so they don't appear in the admin by default. I can change a setting to show them.

akmiller01:
Hey @edwinmp, the Timestamp should show in the admin panel now. Feel free to merge this into develop after your next review.

edwinmp (approved):

PR #59, opened 2019-03-05
Feature/enable rest api

akmiller01:
Dependent on #58 since all authenticated models require a user field. Enables REST api endpoints for viewing Tags and Users.

E.g. `/api/users`
image

`/api/users/1`
image

`/api/tags/`
image

And authentication as demonstrated in the command line:
```
~$ http DELETE http://127.0.0.1:8000/api/tags/1/
HTTP/1.1 403 Forbidden
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Length: 63
Content-Type: application/json
Date: Tue, 05 Mar 2019 16:08:22 GMT
Server: WSGIServer/0.2 CPython/3.6.7
Vary: Accept
X-Frame-Options: SAMEORIGIN

{
    "detail": "You do not have permission to perform this action."
}
```
```
http -a alex:*********** DELETE http://127.0.0.1:8000/api/tags/1/
HTTP/1.1 204 No Content
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Length: 0
Date: Tue, 05 Mar 2019 16:08:50 GMT
Server: WSGIServer/0.2 CPython/3.6.7
Vary: Accept
X-Frame-Options: SAMEORIGIN
```

Needs tests.


edwinmp:
@akmiller01 I hope all these endpoints are built to require an API token. They shouldn't work without it, not even for fetching data.

akmiller01:
Yep @edwinmp should be a simple change in settings.py https://github.com/devinit/ddw-analyst-ui/pull/59/files#diff-1e3260587452284432a23f27fe15da83R52

akmiller01:
Coded up all of the Source related models and added them to a nested field serializer for you @edwinmp, hopefully this will be all you need to get started on the frontend source work.

Here's an e.g. of the nested fields on the Source List:
image

With `sourcecolumnmap_set` and `updatehistory_set` being nested.

And here's an e.g. of the User endpoint now:
image

core/serializers.py
```diff
@@ -0,0 +1,44 @@
...
+        model = SourceColumnMap
+        fields = ('source', 'name', 'description', 'source_name')
+
+
+class UpdateHistorySerializer(serializers.ModelSerializer):
+    source = serializers.ReadOnlyField(source='source.indicator')
+
+    class Meta:
+        model = UpdateHistory
+        fields = ('source', 'history_table', 'is_major_release', 'released_on', 'release_description', 'invalidated_on', 'invalidation_description')
```
edwinmp: Long line... could you move each option to it's own line. More readable. There's a JS/TS linter rule to keep the line length to max 120.

core/serializers.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class SourceSerializer(serializers.ModelSerializer):
+    sourcecolumnmap_set = SourceColumnMapSerializer(many=True, read_only=True)
+    updatehistory_set = UpdateHistorySerializer(many=True, read_only=True)
+
+    class Meta:
+        model = Source
+        fields = ('indicator', 'indicator_acronym', 'source', 'source_acronym', 'source_url', 'download_path', 'last_updated_on', 'storage_type', 'active_mirror_name', 'description', 'user', 'created_on', 'updated_on', 'sourcecolumnmap_set', 'updatehistory_set')
```
edwinmp: This one is even longer...

core/urls.py
```diff
@@ -1,10 +1,18 @@
...
 from knox import views as knox_views
 from core.views import LoginView
+from core import views as core_views
 
 urlpatterns = [
     path('auth/login/', LoginView.as_view(), name='knox_login'),
     path(r'auth/logout/', knox_views.LogoutView.as_view(), name='knox_logout'),
     path(r'auth/logoutall/', knox_views.LogoutAllView.as_view(), name='knox_logoutall'),
     path(r'auth/', include('rest_social_auth.urls_knox')),
+    path(r'rest/', include('rest_framework.urls')),
```
edwinmp: As far as I can tell, this one adds login & logout urls for the DRF. Do we need those? When I logged in, I got the error below

image

core/urls.py
```diff
@@ -1,10 +1,18 @@
...
+    path('users/', core_views.UserList.as_view()),
+    path('users/<int:pk>/', core_views.UserDetail.as_view()),
+    path('tags/', core_views.TagList.as_view()),
+    path('tags/<int:pk>/', core_views.TagDetail.as_view()),
+    path('sources/', core_views.SourceList.as_view()),
+    path('sourcess/<int:pk>/', core_views.SourceDetail.as_view()),
```
edwinmp: `sources` ... typo

core/views.py
```diff
@@ -1,7 +1,65 @@
...
 from rest_framework.authentication import BasicAuthentication
+from core.models import Tag, Source
+from core.serializers import TagSerializer, UserSerializer, SourceSerializer
+from rest_framework import generics, permissions
+from django.contrib.auth.models import User
+from core.permissions import IsOwnerOrReadOnly
+
+
+class UserList(generics.ListAPIView):
+    authentication_classes = [BasicAuthentication]
```
edwinmp: Change all these to use `TokenAuthentication`... can't be entering user credentials for every request.

`from knox.auth import TokenAuthentication`

edwinmp (changes requested):
Good. Could you add support for PUT & DELETE, cos we're going to have to perform all CRUD operations from the FE via the API.

core/serializers.py
```diff
@@ -0,0 +1,44 @@
...
+from core.models import Tag, Source, SourceColumnMap, UpdateHistory
+from django.contrib.auth.models import User
+
+
+class TagSerializer(serializers.ModelSerializer):
+    user = serializers.ReadOnlyField(source='user.username')
+
+    class Meta:
+        model = Tag
+        fields = ('name', 'user', 'created_on', 'updated_on')
```
edwinmp: Add the `id` field as well.

core/serializers.py
```diff
@@ -0,0 +1,44 @@
...
+        model = User
+        fields = ('id', 'username', 'tag_set')
+
+
+class SourceColumnMapSerializer(serializers.ModelSerializer):
+    source = serializers.ReadOnlyField(source='source.indicator')
+
+    class Meta:
+        model = SourceColumnMap
+        fields = ('source', 'name', 'description', 'source_name')
```
edwinmp: Here too.

core/serializers.py
edwinmp: Add `id` here as well.

edwinmp (changes requested):

core/permissions.py
```diff
@@ -0,0 +1,16 @@
...
+    Custom permission to only allow owners of an object to edit it.
+    """
+
+    def has_object_permission(self, request, view, obj):
+        # Read permissions are allowed to any request,
+        # so we'll always allow GET, HEAD or OPTIONS requests.
+        if request.method in permissions.SAFE_METHODS:
+            return True
+
+        # Write permissions are only allowed to the owner of the snippet.
```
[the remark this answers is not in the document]
akmiller01: Hey @edwinmp , I couldn't reply to your top review, but PUT and DELETE are allowed at the `Detail` level by owners right now. I'll add it for superusers as well.

core/permissions.py
```diff
@@ -0,0 +1,16 @@
+from rest_framework import permissions
+
+
```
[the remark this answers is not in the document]
akmiller01: Python linting requires 2 newlines before a Class.

akmiller01:
Okay, so removing the REST Framework urls has removed the login button from the View GUI. Is that what we want? 
image

akmiller01:
Also @edwinmp I saw you setup `/api/auth/login` View with `BasicAuthentication`. Do you want to change that to Token as well?

edwinmp:
@akmiller01 while the View GUI looks good, we have no intention of using it. I suggest you use a nice rest client like `Postman`.

As for the `/api/auth/login`, this is the only path that should require basic authentication. It returns the token that's used in all other API requests.

Header is `Authorization token [token_goes_here]`

core/permissions.py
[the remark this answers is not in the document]
edwinmp: Ah, so it seems... had never noticed before.

Napho:
@akmiller01  from what i understand from @edwinmp , we need to use JWT. Might require additional models. Please make the push and i can look at that

edwinmp (approved):

PR #61, opened 2019-03-08
Feature/db router

akmiller01:
Hey Edwin and Naphlin,

Worked on this Friday evening. It is complete, but will definitely clash a bit with `feature/pypika` so we need to make sure we resolve those once one gets merged (I'd prefer this one first).

I made a separate app for the dataset models, and Django seems to like them and even displays records in the Admin interface if we want. The only issue is we cannot test that new app, because all of the models are unmanaged. So Django tries to spin up a new DB to test them with, and then fails because the new DB doesn't have any tables.

Not sure we'll ever need to access the database models like this, but this router will make even direct connections easier.

```
>>> from data.models import CrsCurrent
>>> cc = CrsCurrent.objects.first()
>>> cc.year
1973
```


edwinmp (commented):
Only issue is that the pull request is breaking on Travis. Guess that's what you were explaining up there... any solution?

akmiller01:
Yep, I'm fixing that now. But I also realized our fixture isn't loading right so I'm working on that as well...

edwinmp (approved):

PR #62, opened 2019-03-10
feature/ui/data sources

edwinmp:
- [x] View list of data sources
- [x] View data source details i.e. metadata, columns, update history
- [x] Search through data sources on Home Page


edwinmp:
@akmiller01 I see no direct connection between the `Theme`, `Sector` models and the `Source`, so I think I may have misunderstood their purpose. `Theme` has a connection to `Query`, which I'm also not sure about. Could you please clarify.
My thinking was that themes and sectors group data sources (`stats.oecd` style), or will that be done exclusively through tags? Don't think tags should be used that way.
I need a construct that can facilitate a tree view similar to the one on `stats.oecd`, but if our data doesn't fall into many disparate themes/sectors, then there's no need for it. Can use the view below

image

akmiller01:
I think that's a question to involve @Napho on as well! The models so far have been based on his initial SQL work. What was your original conception behind themes and sectors being on Operations, and not on sources themselves @Napho ?

Napho:
@edwinmp the way i reasoned it is that sources are just sources, they have a huge chunk of data without a lot of direction. For you to get any meaning out of it then you will have to do some filtering or transformation or joins etcetera.. A group of these  is what makes an operation; the operations target certain sectors that DI analysts work on eg Development Activities, Aid, Education etc and a theme is what groups operations (you can relate to ddh here .... unbundling aid, international official finance, imf relief funds, pneumonia etc)

edwinmp:
@Napho Can a source hold data that falls under more than one theme/sector?

Napho:
> @Napho Can a source hold data that falls under more than one theme/sector?

Yes

Napho:
Just one comment, on the table columns section, i think it would be good to show both the translated name and the source name side by side, not sure which one you are going to showing though. otherwise LGTM!! 
Screen Shot 2019-03-14 at 12 45 26

Napho (approved):

PR #75, opened 2019-04-05
feature/query-builder/select

edwinmp:
Handles the frontend select operation for the query builder


k8hughes:
On the Query Step - http://138.68.138.212/queries/build/ a user should be able to save.

Remove "sample output path" field.

When someone clicks save and preview they should be able to go back to the query builder.

On home: you should be able to click the name of the query to open it.  You should also be able to export the data from the home / saved query view.

Duncan-Knox:
Option to remove query on the home page. Maybe under 'actions'.

akmiller01:
Perhaps copy query button for when we have user permissions?

edwinmp:
`Edit Operation` button on Query Data view

akmiller01:
We might need to rethink the audit log. It looks like POSTs aren't recording the user faithfully:

image

akmiller01:
@edwinmp adding more sources now. Looks like we could use a "Next page" button for sources, as well as removing "()" when no acronym exists.

image

edwinmp:
@akmiller01 yeah, on the View Data page as well... postponed adding pagination though... Will do that as part of clean up and enhancements. Made a note of it. Good catch on the acronyms... 👍

akmiller01 (approved):
Working well! :shipit:

Issue #41, opened 2019-02-15
Export


k8hughes:

Duncan-Knox:
I am getting instances of 'download failed - Network error'. This appears to be from attempts to export larger files. Small size files seem to export fine. Is there a limit here and can this be altered possibly?

Duncan-Knox:
Still having the same issue as above with larger files. This is the same instances of 'download failed - Network error'.  Has the fix come through yet?

Duncan-Knox:
Definite improvement on the file size download but still seeing some larger files not being returned. The view data pane spends a while loading before stating 'No results found', this is possibly when request is too large or might be do to with something in the data. I'm testing further scenarios when this can be the case (ie more complex query builds). Wondering if we can plan to increase the limits available or they have reached the maximum?

edwinmp:
Hi @Duncan-Knox ... could you please share a link to the data page, or the name of the operation.

Duncan-Knox:
This comes from a position of not knowing what the limits can be. Appreciate there is a point at which exports are too large to be of use and it is of course not reasonable to expect a near limitless download. 

I've added a query 'CRS year extract 02-05-2019' http://138.68.138.212/queries/data/66/. Wondering whether it is possible to extract this size of file?

I'm aiming to continue testing on seeing whether more complex queries (on smaller datasets) end with the 'No results found', and post if I spot anything.

Duncan-Knox:
Definitely would argue knowing any query size limits would be useful information for users, if possible?

Duncan-Knox:
Edit  @edwinmp had accidentally deleted the query previously linked. New data page link here: http://138.68.138.212/queries/data/66/

edwinmp:
@Duncan-Knox @akmiller01  The operation above returned a `502 Bad Gateway` error. Data was returned for limited operations of up to 10,000 records... It broke when I tried 100,000. My guess is that a timeout occurs because the operation takes too long @akmiller01 ... I'm looking into optimisation alternatives  for this, mostly because I think the response shouldn't be this slow or even fail over just under 100,000 records...

akmiller01:
@edwinmp does the branch you tested on include the estimated-count refactor already?

edwinmp:
If it was merged into `develop`, then yeah...

akmiller01:
```
>>> from core.models import Operation
>>> big_query = Operation.objects.get(pk=66)
>>> big_query.build_query(estimate_count=True)
[...]

Issue #39, opened 2019-02-15
Aggregate


k8hughes:
Select columns, select aggregation function, select column

Duncan-Knox:
Under the operation steps the 'Of' box does not return any columns to select or allow anything to be typed and entered to select a column. It states 'no results found'. Is this a bug?

edwinmp:
@akmiller01, are you adding the `data_type` field to the columns? Won't be able to show any columns in the drop-down without it.

akmiller01:
Yep. It's already in my `models.py` `SourceColumnMap` object. Once I get Duncan's last piece of data in, we can merge data-collection-folder to develop?

edwinmp:
They all currently seem to have a `data_type` of `null`, so if the merge can fix that, then definitely merge.

edwinmp:
Hi @Duncan-Knox, the column issue should be fixed now ... please note that for arithmetic operations, only columns with numerical data shall be returned.

Duncan-Knox:
Hi @edwinmp thank you for the update. I look forward to testing the aggregation and transform functions.

edwinmp:
Hi @Duncan-Knox ... have you encountered any more issues with the aggregate or transform? If not, I'd like to close this.

Duncan-Knox:
Hi @edwinmp , Let me cover the full range of functions now. I've not tested all within aggregate.

Duncan-Knox:
I have not encountered any further bugs testing the aggregate function (beyond that raised on distinct). 

A thought on standard deviation. I'm tempted to say we note that the standard deviation operation refers to a sample standard deviation as opposed to population so users are clear, ie title 'standard deviation (sample)'. This said, for most of the data within the UI at the moment, I'd expect to apply a population SD (complete data at the country level). Also bear in-mind, in my experience I'd guess SD would be a relatively low use function (occasional internal work). How does this square with your experience @akmiller01?

akmiller01:
@Duncan-Knox true. Maybe the likes of Zach and Dan W would use STDDEV more frequently.

k8hughes:
@Duncan-Knox can this be closed or is more work required?

Duncan-Knox:
Done @akmiller01 and @k8hughes. Checked with Dan.W and he uses STDDEV for occasional internal work. Both sample and population. Feel free to add 'sample' in the title in the case you think it may be useful to make that clear to users or is necessary.

Duncan-Knox:
[...]

Issue #37, opened 2019-02-15
Calculate


k8hughes:
Select datasource, select columns, select transformation

The step by step element for the implementation of this probably should be sketched and share with the team so that we are all in agreement.  @akmiller01 had a clear idea of what the user needs here.

Duncan-Knox:
Similar to issue raised here: https://github.com/devinit/ddw-analyst-ui/issues/39.

Scalar transform and Multi transform do not allow the user to select any columns under the 'On' box.

Duncan-Knox:
Scalar transform options Add, Multiply, Power Subtract and Divide run smoothly. 

Option Concatanate appears to Concatanate with the title of the column selected under 'on' as opposed to the value on a single row. IE typing 'X' to concatanate with a column called 'ISO3' will mean each cell in newly formed 'ISO3 concat' will show as 'ISO3 X' as opposed to 'AFG X' .. 'BHU X' and so on. Flagging as this is quite possibly a bug?

Was also wondering if the concatanate function would have an aim to be developed to allow a merge between multiple columns as well as manually added text?

akmiller01:
That's the intended behavior for concatenate in it's scalar form. We would need to add a multi transform version of concat to concatenate multiple columns together.

Duncan-Knox:
Noted on the multiple columns concat being something that would not go on scalar and would be part of multi transform instead. Not a request as such for the time being.

Is scalar transform concat merging with the column title intended behaviour? I ask as the numerical scalars are applied to the row values within columns. Guess the same logic may have been in place for concat, hence the comment, unless I'm missing something?

akmiller01:
Ah okay, no I misread your message. Every row shows as "ISO3 X" for e.g.?

Duncan-Knox:
erm like this: 

![image](https://user-images.githubusercontent.com/13347058/57026467-cd980480-6c31-11e9-9c14-e7f1a95347a8.png)

Wondering if that was intentional and if so what the logic was?

Duncan-Knox:
As would have guessed it would show 'AFGX' and so on, instead of 'ISO3X' if that makes sense?

akmiller01:
Yep, that is a bug. Working on a fix.

akmiller01:
Will be here https://github.com/devinit/ddw-analyst-ui/pull/90

Duncan-Knox:
Ah right great, noted.

akmiller01:
Fixed in the PR 
[...]

Issue #36, opened 2019-02-15
Join


k8hughes:
Select datasource, select columns, specify joins (left, right, e.t.c)

Duncan-Knox:
Hi @edwinmp and @akmiller01. Join has been fantastic to trial and very smooth. One possible bug -   I'm not sure the full range of selection options for queries are available after the joins have been made. For example, if I join dataset 'non-grant government revenue' to  'states of fragility 2018' I won't get the option to perform queries (aggregate, filter, etc) based on the 'states of fragility 2018' columns. Is this one to look into?

Duncan-Knox:
At times the options for 'Data Set To Join With' will only show data sets to join that are on page 1 (1-10 of 36) of datasets. Not quite sure what drives this but appears to be the case once queries become more complex.

akmiller01:
Probably the same bug that was preventing pagination on the data sources in
the first place.

Duncan-Knox:
Possible bug/regression has cropped up on the Join Query. It states  'At least one column mapping is required' above the two columns selected to for joining. This appears to occur when two matching columns have been selected.

edwinmp:
Hi @Duncan-Knox ... the above issues on join should be fixed on staging.

Duncan-Knox:
Hi @edwinmp, sorry, I still think the most recently posted join bug on 'At least one column mapping is required' is occurring when two matching columns are selected. Is it a case of waiting a little longer for the fix to show?

edwinmp:
@Duncan-Knox Could you please clear you cache or try incognito....

Duncan-Knox:
Sure, @edwinmp was working on incognito but turned out that a new session was needed. The column mapping bug fix has gone through ok.

akmiller01:
Hmm... It really shouldn't default to removing NA. That's really R-exclusive behavior, so it may be missing due to the import routine in R rather than the join. What dataset was missing Namibia @Duncan-Knox ?

Duncan-Knox:
Ah the ISO2 has disappeared. The rest of the row is available for Namibia with blank ISO2. Hence the disappearance after merging. That was in 'DI Deflators Constant 2017 USD' but possibly others too @akmiller01

Duncan-Knox:
I think this occurs in any dataset with ISO2 for NA that has been manually put together.

akmiller01:
[...]

PR #77, opened 2019-04-11
Feature/query builder/aggregate

edwinmp:
Contains:

- [New] Aggregate (issue #39)
- [New] Transform (Scalar & Multi) - issue #44 #37  
- [New] Join - issue #36 
- [Update] add pagination to data sources
- [Update] add pagination to operation data
- [New] operations that result in custom columns, these columns are always shown in the results
- [Update] add pagination to data sources
- [New] add Export to CSV for operation data - issue #41 
- [Update] format original column names for use as captions
- [New] reset button on query builder


Duncan-Knox:
Adding secondary filter on 'create operation step' to an existing operation with a filter seems to function differently to adding a (second) filter to a new operation step. The former does not seem to apply both filters. The latter does, which is preferable as it allows the user to use two filters to create a between operation. This is not a significant issue but may cause confusion.

Duncan-Knox:
Query Data: is it possible at all to increase the number of columns the user can view on 'view data'?

edwinmp:
> Adding secondary filter on 'create operation step' to an existing operation with a filter seems to function differently to adding a (second) filter to a new operation step. The former does not seem to apply both filters. The latter does, which is preferable as it allows the user to use two filters to create a between operation. This is not a significant issue but may cause confusion.

Correct me if I'm wrong, @akmiller01 ... multiple filter options on the same step translate as an `OR` operation (rows that match one of them are returned), while adding multiple filter steps on an operation translates as an `AND` (rows that match both are returned). We'll need to add documentation/help that explains this.

edwinmp:
> Query Data: is it possible at all to increase the number of columns the user can view on 'view data'?

Not at the moment. They're hardcoded to 6. Now that I've added the `Select`, I'll show all columns. Will be up to the user to use the `Select` operation to limit how many are shown.

Duncan-Knox:
> > Adding secondary filter on 'create operation step' to an existing operation with a filter seems to function differently to adding a (second) filter to a new operation step. The former does not seem to apply both filters. The latter does, which is preferable as it allows the user to use two filters to create a between operation. This is not a significant issue but may cause confusion.
> 
> Correct me if I'm wrong, @akmiller01 ... multiple filter options on the same step translate as an `OR` operation (rows that match one of them are returned), while adding multiple filter steps on an operation translates as an `AND` (rows that match both are returned). We'll need to add documentation/help that explains this.

Thanks @edwinmp , I should have seen it this way. This makes sense and sounds very useful to have this set-up.

akmiller01 (approved):
Bugs seem to be fixed!

PR #86, opened 2019-04-29
Feature/append isos

akmiller01:
Small feature to join ISO table to FTS upon import


edwinmp:
@Napho could you please review this ...

Napho (approved):

PR #87, opened 2019-04-29
Much faster estimated count

akmiller01:
Cuts the amount of time for a small query in about half:
```
time python manage.py shell -c 'from core.models import Operation; one=Operation.objects.get(pk=1); print(one.query_table(1, 1, True));'
real	0m0.667s
user	0m0.506s
sys	0m0.058s

time python manage.py shell -c 'from core.models import Operation; one=Operation.objects.get(pk=1); print(one.query_table(1, 1, False));'
real	0m1.339s
user	0m0.496s
sys	0m0.055s

```

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -101,16 +101,16 @@ class Operation(BaseEntity):
     def __str__(self):
         return self.name
 
-    def build_query(self, limit=None, offset=None):
+    def build_query(self, limit=None, offset=None, estimate_count=None):
         """Build an SQL query"""
-        count_query = QueryBuilder(self).count_sql()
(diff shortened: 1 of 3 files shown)


akmiller01:
Also yields wildly inaccurate count results... But could be phrased as "Maximum possible count."

In the example above, real count is 3,095 and estimate just yields the full count of the table which is 3,313,686.

Napho (approved):
LGTM

akmiller01:
Would this have any conflict with your front-end @edwinmp ? It still should return a single number for "count".

PR #91, opened 2019-05-02
Fix/psycopg2 memleak

akmiller01:
Part fix, part feature, part proof-of-concept. Feel free to change this how you guys see fit @edwinmp or @Napho . It may make more sense to try and stream this to an HTTP response than a static file, but this is the bare minimum backend code to get big datasets out. This code is on staging now if you want to play with it in Django shell

```
root@ddw-analyst-ui-dev:~/ddw-analyst-ui# docker-compose exec web python3 manage.py shell
>>> from core.models import Operation
>>> big_op = Operation.objects.get(pk=66)
>>> big_op.export_data()
(257430, '_django_curs_139966473828096_1.csv')
```

Saves into the `static` folder a randomly named massive 213 MB CSV. NGINX then servers that static folder, so you can download the file here http://138.68.138.212/static/_django_curs_139966473828096_1.csv

The name is generated off of the cursor name, which is randomly generated to stay thread-safe.

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -5,7 +5,7 @@ from django.contrib.auth.models import User
 from django.db import models
 
 from core.pypika_utils import QueryBuilder
-from data.db_manager import fetch_data
+from data.db_manager import fetch_data, stream_to_file
 
 # from core.models_template import *
 
@@ -110,7 +110,10 @@ class Operation(BaseEntity):
 
     def query_table(self, limit, offset, estimate_count):
         """Build a query then execute it to return the matching data"""
-        return fetch_data(self.build_query(limit, offset, estimate_count))
+        queries = self.build_query(limit, offset, estimate_count)
+        if limit is not None and limit > 10000:
+            return fetch_data(queries)
+        return stream_to_file(queries)
 
 
 class OperationStep(BaseEntity):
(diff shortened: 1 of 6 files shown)


akmiller01:
Okay, turned this proof of concept into an HTTP Stream. Completely scrapped the stream-to-file function. Right now the view is written as an API view because I wasn't able to figure out the Knox auth without doing so. Are you able to incorporate an APIView data-stream into your front-end @edwinmp ? Here's what it looks like in Postman:

image

Napho (dismissed):

akmiller01:
Okay @edwinmp so I spent several hours on this and failed miserably. I thought for a moment I would be able to override KnoxLoginView's post method to authenticate the user on the Python-server-side when the frontend posts the credentials on login... But of course the browser never gets the cookie from that POST so that didn't work... And then I tried to grab USER and API_KEY from LocalForage through the `request.session` in Python and of course that didn't work. I don't think it's going to be possible to POST the auth from the JS side to the Python side without actually making some front-end code to achieve that. And the only other alternative is hiding the export behind BasicAuthentication which would require a new login each time an export is made...

TLDR; I reverted my code back to the point where it's acting like an API endpoint expecting the token authentication, and I hope you can make a front-end wrapper around it @edwinmp

edwinmp:
@akmiller01 may as well leave it as is. Important thing is being able to stream data to a file. Besides, this is read-only. The security is mostly aimed at preventing unauthorised write or delete operations.

akmiller01:
Tested and working form-data POST with token:

image

edwinmp (approved):

PR #94, opened 2019-05-03
Don't drop Namibia ISO2

akmiller01:
image


Duncan-Knox (approved):
Changes approved. Will perform a quick check on NA ISO values in the UI once this has gone through.

PR #96, opened 2019-05-07
Fix/pypika joins

akmiller01:
Fixes Pypika step ordering

Changes join behavior to alias columns in common between Table1 and Table2.

### DIFF
--- a/core/pypika_utils.py
+++ b/core/pypika_utils.py
@@ -188,13 +188,10 @@ class QueryBuilder:
         unjoined_table1_columns = [col for col in columns_x if col not in joined_table1_columns]
         unjoined_table2_columns = [col for col in columns_y if col not in joined_table2_columns]
         common_unjoined_columns = list(set(unjoined_table1_columns) & set(unjoined_table2_columns))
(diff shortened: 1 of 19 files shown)


edwinmp:
@akmiller01 added the FE changes for this...

akmiller01:
@edwinmp just gave this a test. Are you able to add to the dropdown the columns which result from a join?

For e.g. if you take CRS and join CRS_isos, you would expect in the next step that ISO3 would be a column on the left-hand dropdown for future joins:

image

In other words, the new state of the columns should be the union of the columns from previous joins.

edwinmp:
@akmiller01 not yet ... I'm not yet executing each step as it's saved. The screenshot is showing an old view of the join operation though... please pull.

akmiller01:
You don't need to execute a step to know the columns from a join. You could store the state of the columns at any given step as the union of all joined datasets columns up to that point http://2ality.com/2015/01/es6-set-operations.html

akmiller01:
I think this is ready to merge @edwinmp Just did a pretty complicated test.

1. Join ISOS to CRS
2. Aggregate ISOS by CRS and Year
3. Join aggregated CRS to country income groups by ISO and year

Worked flawlessly

image

edwinmp (approved):

PR #100, opened 2019-05-13
Feature/test queries

akmiller01:
Hey @edwinmp and @Napho , I got fed up with CI missing a way to actually check database output, so I brute-forced out this method of getting our dataset fixtures to un-unmanage themselves during tests. Added two quick output tests, but I'm hoping this is a model we can build on to prevent needing @Duncan-Knox to constantly manually check some of our core features.

That being said, the biggest caveat here is that we can only test on data in the fixtures. So this won't be useful for checking particular datasets, but ensuring that our PSQL actually returns expected output with known data.


akmiller01:
Already paying testing dividends... Median doesn't seem to work `function median(numeric) does not exist`

And this also picked up on the order of operations problem fixed in `fix/pypika-joins.` But I'll need to wait for that to be merged into `develop` before incorporating the fix for `test_can_sum_from_joined_column_passes`

edwinmp (approved):

PR #101, opened 2019-05-14
Err handling db manager

akmiller01:
Really simple error handling with try/catch on TypeError and ProgrammingError. Handles the err thrown by bad SQL as well as the err thrown by zip on empty results. E.g. Operation 7 requests a missing column, and Operation 8 yields an empty dataset.

```
>>> from core.models import Operation
>>> op = Operation.objects.get(pk=7)
>>> op.query_table(1, 0, False)
(1, [{'error': 'column "iso10" does not exist\nLINE 1: ...ountry_name","di_id","year","income_group","iso3","iso10" FR...\n                                                             ^\nHINT:  Perhaps you meant to reference the column "historical_income_groupings.iso3".\n'}])
>>> op = Operation.objects.get(pk=8)
>>> op.query_table(1, 0, False)
(0, [])
```

### DIFF
--- a/core/serializers.py
+++ b/core/serializers.py
@@ -5,6 +5,7 @@
     Serializers also provide deserialization, allowing parsed data to be converted back into complex
     types, after first validating the incoming data.
 """
+from json.decoder import JSONDecodeError
 from django.contrib.auth.models import Permission, User
 from rest_framework import serializers
 from rest_framework import pagination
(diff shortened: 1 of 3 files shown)


akmiller01:
Once #100 is merged in, I can add a test for this.

akmiller01:
Two tests for the two error cases added in @edwinmp , this branch is ready for review and merge.

edwinmp:
@akmiller01 was running into a JSON error on an old operation, so I added handling for that as well.

edwinmp (approved):

PR #115, opened 2019-05-21
Feature/ssl

akmiller01:
All the settings for SSL. You just need to make an ssl folder and add your `fullchain.pem` and `privkey.pem` files in there. For testing with self-signed certs:

```
mkdir ssl
cd ssl
openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout privkey.pem -out fullchain.pem
```
Or for continuing local development as if nothing has changed:
```
export DJANGO_DEV=True
```

Also, in prod we'll want to
```
export DJANGO_SECRET_KEY=an-actually-secret-key
```

Also contains proxy_buffering code from #114 so I'm just closing that.

### DIFF
--- a/config/nginx/django.conf
+++ b/config/nginx/django.conf
@@ -6,6 +6,8 @@ upstream web {
 server {
     listen 80;
 
+    server_name ddw.devinit.org;
+
     client_max_body_size 10M;
 
     location /static/ {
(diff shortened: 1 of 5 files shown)


config/nginx/django.conf
```diff
@@ -6,6 +6,39 @@ upstream web {
 server {
```
Napho, line 6: @akmiller01 
shouldn't this redirect to https then?

Napho (changes requested):

config/nginx/django.conf
akmiller01: It upsets Django a bit if you make Nginx do the redirect instead of WSGI. So the SECURE_SSL_REDIRECT is set in settings.py

Napho: cool

Napho (approved):

Issue #93, opened 2019-05-03
Firefox export csv bug


k8hughes:
When using firefox I have discovered that the esport cvs button doesn't work.
I discovered this when looking at the already saved queries for example this page:  http://138.68.138.212/queries/data/31 

No error code is showing in the console.

If this is hard to fix we could leave it out for now and tell users to start using hte app on Crome initially.  (not launch blocking)

edwinmp:
@k8hughes I cannot replicate this. What's the Firefox version you're using? Once you take note of that, please try to update the browser (if it's not updated already) and test again.

edwinmp:
I'm closing this since I cannot replicate it in the latest version of Firefox.

PR #117, opened 2019-05-23
Patch/1.0.2

edwinmp:
- [x] Year 'is equals to' should be 'is equal to'. 'Is greater than or equals to', etc.
- [ ] All modern browser compatibility (#93)


akmiller01 (approved):
:+1:

PR #126, opened 2019-05-31
IATI Datastore mirror

akmiller01:
Might need more transformation, or more metadata to be useful. Right now the numerical columns vary by currency, so we might need a currency conversion rate table in the future.


edwinmp (dismissed):

akmiller01:
Had to rebase. Mind re-reviewing @edwinmp ?

edwinmp (approved):

PR #143, opened 2019-10-10
Reference files, as requested by Duncan

akmiller01:


edwinmp (dismissed):

akmiller01:
Ah. We must have a requirement that CSVs are named for their tables. I'll rename
```
(Pdb) active_mirror_name
'OECD CRS aid type reference file'
```

edwinmp (approved):

Issue #83, opened 2019-04-26
WEO


Duncan-Knox:
Hi @akmiller01 another minor name change suggestion: 'World Economic Indicators' to 'World Economic Outlook'?

akmiller01:
Fixed in code!

PR #148, opened 2020-04-28
Feature/april2020data

akmiller01:


akmiller01:
Hi @k8hughes , I can't find an issue for it, but it looks like we were generally talking about updating the WEO around April this year: https://github.com/devinit/ddw-analyst-ui/issues/83

Yes, this needs to be merged if we want the latest WEO data in the database.

Duncan-Knox:
I don't recall any decisions from our side on what to do here. The IMF outlined their changes here:
[https://www.imf.org/external/pubs/ft/weo/2020/01/weodata/index.aspx](url)

'Due to the high level of uncertainty in current global economic conditions, the April 2020 WEO database and statistical tables contain only these indicators: real GDP growth, consumer price index, current account balance, unemployment, per capita GDP growth, and fiscal balance. Projections for these indicators are provided only through 2021.'

I don't think the old predictions are usable any more but the fact they've taken away a lot of the indicators we used historical data for is not ideal. There's merit in both updating and keeping on hold.

k8hughes:
It would be good to have a decision on what to do, I think if the old data predictive data we hold is now not accurate then we should remove it.  @Duncan-Knox who else uses this data that we should ask for their view?

Duncan-Knox:
It is fairly general dataset so no one specifically comes to mind as a key user. I'd vote for updating it with the new data because 
it represents a newer release and as you say, more accurate compared to the previous release.

The older data is still accessible on the IMF website here if anyone needs it. We could make this clear to users if we choose to do the update? https://www.imf.org/external/pubs/ft/weo/2019/02/weodata/index.aspx

akmiller01:
Agreed with Duncan. It probably does make sense to update.

wakibi:
Not exactly related to this PR but @k8hughes could this be one of those times when #185 comes in handy?

k8hughes:
@wakibi simialr but not quite the same.  If we're agreed this update can go forward then it would be good to merge.

k8hughes:
@akmiller01 what are we doing with this PR in the end?  Do we need it?

akmiller01:
It's been subsumed by the Oct 2020 WEO. So we can close it.

Issue #158, opened 2020-05-13
Create Django models for managing scheduled events


edwinmp:
Take inspiration from [here](https://docs.mendix.com/refguide/scheduled-events)
- Scheduled event CRUD
- Scheduled event audit trail & history CRU

**Models** (inherit BaseEntity)
_ScheduledEvent_
- Name
- Description
- Script Name
- Enabled
- Start DateTime
- Repeat (Bool)
- Interval (number)
- Interval Type (minutes, seconds, hours, days, weeks, months, years)

_ScheduledEventRunInstance_
- ScheduledEvent (ForeignKey)
- StartedAt (DateTime)
- EndedAt(DateTime)
- Status (Pending, Running, Completed, Errored)

PR #169, opened 2020-05-18
Feature/scheduled event model

davidebukali:
Feature for #158


core/admin.py
```diff
@@ -1,5 +1,5 @@
 from django.contrib import admin
-from core.models import Sector, Theme, Tag, Operation, OperationStep, Review, Source, SourceColumnMap, UpdateHistory, AuditLogEntry
+from core.models import Sector, Theme, Tag, Operation, OperationStep, Review, Source, SourceColumnMap, UpdateHistory, AuditLogEntry, ScheduledEvent, ScheduledEventRunInstance
```
edwinmp: This looks very long ... implement a multi-line alternative.

core/admin.py
```diff
@@ -16,3 +16,5 @@ class AuditLogEntryAdmin(admin.ModelAdmin):
 admin.site.register(SourceColumnMap)
 admin.site.register(UpdateHistory)
 admin.site.register(AuditLogEntry, AuditLogEntryAdmin)
+admin.site.register(ScheduledEvent)
+admin.site.register(ScheduledEventRunInstance)
```
edwinmp, line 34: Excellent.

core/models.py
```diff
@@ -194,3 +194,41 @@ def __str__(self):
...
+        ('hrs', 'Hours'),
+        ('dys', 'Days'),
+        ('wks', 'Weeks'),
+        ('mnt', 'Months'),
+        ('yrs', 'Years'),
+    )
+    name = models.TextField(null=False, blank=False)
+    description = models.TextField(null=False, blank=True)
+    script_name = models.TextField(null=False, blank=False)
+    enabled = models.BooleanField()
```
edwinmp: Give it a default = True

core/models.py
```diff
@@ -194,3 +194,41 @@ def __str__(self):
...
+    start_date = models.DateTimeField(null=False, blank=False)
+    repeat = models.BooleanField()
```
edwinmp: Default = False

core/models.py
```diff
@@ -194,3 +194,41 @@ def __str__(self):
...
+    interval = models.BigIntegerField(blank=True, null=True)
+    interval_type = models.CharField(max_length=1, choices=interval_type_choices, null=True, blank=True)
```
edwinmp: hmm... confused ... max_length=1 for a CharField? And this is working with your choices? Cos all your choices have at least 3 characters...

core/models.py
```diff
@@ -194,3 +194,41 @@ def __str__(self):
...
+    """Scheduled Event Run Instances."""
+    status_choices = (
+        ('p', 'Pending'),
+        ('r', 'Running'),
+        ('c', 'Completed'),
+        ('e', 'Erroed'),
+    )
+    scheduled_event = models.ForeignKey(ScheduledEvent, on_delete=models.CASCADE)
+    start_at = models.DateTimeField(null=False, blank=False)
+    ended_at = models.DateTimeField(null=False, blank=False)
```
edwinmp: This can be blank ... cos the instance shall be updated with this value after the task has finished running. So it'll need to be initially saved without an `ended_at`

edwinmp (changes requested):

core/admin.py
davidebukali: Updated

core/models.py
davidebukali: Added default value

core/models.py
davidebukali: Updated

core/models.py
davidebukali: It should be at most 3, now updated

core/models.py
davidebukali: Updated

edwinmp (approved):

Issue #166, opened 2020-05-13
Create a wizard for updating select data tables


edwinmp:
- A wizard that operates similarly to https://app.datawrapper.de/
- Initially accepts CSV & Excel as inputs
- [Low Priority] accepts API as input

edwinmp:
Currently waiting on the backend dependency in issue #165

PR #172, opened 2020-05-21
Feature | update wizard design

edwinmp:
Design step for issue #166 

Accessible via http://127.0.0.1:8000/designs/update-data-source/


akmiller01:
Looks great @edwinmp I do wonder if we need the `Map` step right now given the fact that we're only updating a preset list of existing tables?

akmiller01 (approved):

edwinmp:
> Looks great @edwinmp I do wonder if we need the `Map` step right now given the fact that we're only updating a preset list of existing tables?

Can't see how we'll point XLS/CVS columns to data table columns without that ... I'd rather not have to hardcode CSV column header names and have to keep telling the analysts what they are for every table ... the mapping will be pretty basic

akmiller01:
> > Looks great @edwinmp I do wonder if we need the `Map` step right now given the fact that we're only updating a preset list of existing tables?
> 
> Can't see how we'll point XLS/CVS columns to data table columns without that ... I'd rather not have to hardcode CSV column header names and have to keep telling the analysts what they are for every table ... the mapping will be pretty basic

Good point!

Issue #161, opened 2020-05-13
Manage the creation and triggering of scheduled events


edwinmp:
Dependent on #158
Since these will be limited to the existing commands/scripts we create on the backend, this will be a purely backend task. And we'd like to limit the FE work on this.
- We can use the Django admin to create the scheduled events (populate dependent model)
- Create a management command that finds and triggers events that are due
- Create a cronjob for executing the management command
- Update executable scripts or management commands to periodically update their respective ScheduledEventRunInstance status at different stages of execution

PR #175, opened 2020-05-22
Feature/management_command_for_scheduled_events

davidebukali:
#161 

All the code for this command lives in `data/management/commands/run_schedules.py`

Like all management commands, it starts in the `handle` function and loops through all enabled `ScheduledEvent`

For each `ScheduledEvent` we get all its `ScheduledEventRunInstance` and check if the schedule is due to run.

The check is started from `check_if_schedule_is_due` which returns either `True` to create a run instance or `False` to skip that run.

If the `ScheduledEvent` does not have any `ScheduledEventRunInstance` we return `True` and create a run instance; 

Otherwise if there are created `ScheduledEventRunInstance` we check if any has an `ended_at` column with `null` value; this means there's a run in progress so we return `False` to skip the run.
If all created `ScheduledEventRunInstance` have been ended, then we check if `ScheduledEvent` has repeat enabled using `manage_repeated_schedules` function.

`manage_repeated_schedules`  returns either `True` to run the schedule or `False` to skip it.
If the `ScheduledEvent` has repeat enabled we handle the interval calculation using `check_if_repeat_is_due` function, if repeat is not enabled we return false and skip the schedule run.

 `check_if_repeat_is_due` function loops through all `ScheduledEventRunInstance` to count how many times it appears in the provided interval. For example for `interval` 2 and `interval_type` weeks, it counts how many `ScheduledEventRunInstance` have a `start_at` column that falls in the last two weeks from now.

If the count is less than 2 weeks, it will run the schedule and skip if its more than two


edwinmp:
That was a bit confusing, probably cos it had function names in it ... hehe

Anyway, this is the behaviour I expect.

1. Get all scheduled events that are `enabled`
2. For each scheduled event, check if they have a run instance that has `status = pending` with a start date/time that's less than the current date/time.
3. Before you run each instance, check whether there's a different instance that has a `status = running` - if so, set this one's status to `skipped` and set `ended_at` to current date/time.
4. If none are found in step 3, trigger each run instance that's found on steps 2
5. Within each running script (you'll need to update them - don't think they're that many yet - https://ddw.devinit.org/api/list_update_scripts - at least update the FTS one for staters), check if there are running instances for it that fit the current date/time, and update them at each relevant stage i.e. `running`, `complete`, `errored` e.t.c
6. After running, check if scheduled event is recursive - if so, calculate the next run time and create a new run instance for that future date/time with a `status = pending`

Have I missed anything? Anyway, check with your implementation and see if it fits.

core/models.py
```diff
@@ -227,6 +227,7 @@ class ScheduledEventRunInstance(BaseEntity):
         ('r', 'Running'),
         ('c', 'Completed'),
         ('e', 'Erroed'),
```
akmiller01: Outside of your code, but just spotted this typo. `Errored`

akmiller01 (changes requested):
Looks good, David. And starts up fine on my end. Would you be able to add something to `data/tests.py` to test this? I included a test shell script that just echos here https://github.com/devinit/ddw-analyst-ui/blob/develop/data_updates/test_stream.sh

edwinmp:
@davidebukali what's pending on this? Could you have it all resolved so we can merge it in.

akmiller01:
> @davidebukali what's pending on this? Could you have it all resolved so we can merge it in.

My change requested is for David to add an automated test to `data/tests.py`

akmiller01 (changes requested):

akmiller01 (approved):
Tests passing, looks good!

PR #177, opened 2020-05-26
Feature | update data source

edwinmp:
Fix for #166


akmiller01 (approved):
Looks good. Does it make sense to keep those debug booleans in the view?

edwinmp:
> Looks good. Does it make sense to keep those debug booleans in the view?

Oh, was using those to exclude design pages on production ... the DEBUG also excludes the live reload link in production. I'm open to suggestions on better alternatives.

akmiller01:
> > Looks good. Does it make sense to keep those debug booleans in the view?
> 
> Oh, was using those to exclude design pages on production ... the DEBUG also excludes the live reload link in production. I'm open to suggestions on better alternatives.

Gotcha. That makes perfect sense :+1:

Issue #162, opened 2020-05-13
Manually trigger a scheduled event


edwinmp:
Dependent on #163

PR #188, opened 2020-06-02
Ft view running instances

mariamiah:

PR #189, opened 2020-06-02
Feature/manually trigger schedules

davidebukali:
#162


frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -13,10 +13,12 @@ export const ScheduledEventsTable: FunctionComponent<ScheduledEventTableProps> =
 
   const renderRows = (): ReactNode =>
     props.events.map((event: ScheduledEvent, index: number) => {
+
```
edwinmp: Remove blank line ... maybe even remove the `return` itself :)

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -38,6 +40,7 @@ export const ScheduledEventsTable: FunctionComponent<ScheduledEventTableProps> =
           <th>Enabled</th>
           <th>Start Date</th>
           <th>Interval</th>
+          <th>Run Schedule</th>
```
edwinmp: Name this column `Actions`

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -1,8 +1,13 @@
-import React, { FunctionComponent } from 'react';
-import moment from 'moment';
+import React, { FunctionComponent, useEffect, useState } from 'react';
+import moment, { Moment } from 'moment';
+import { Button } from 'react-bootstrap';
+import * as localForage from 'localforage';
+import { api, localForageKeys } from '../../utils';
+import axios from 'axios';
```
edwinmp: Organise imports

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,52 @@ export interface ScheduledEventsTableRowProps {
...
+  start_at: Moment;
+  status: string;
+}
+
+const BASEPATH = api.routes.VIEW_SCHEDULED_EVENTS;
+
+const createRunInstance = async (
+  scheduleId: number | undefined,
+  data: RunInstancePayload,
+): Promise<any> => {
```
edwinmp: Don't you know the shape this promise's return is going to take? Give it a type ... no `any`

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,52 @@ export interface ScheduledEventsTableRowProps {
...
+    data,
+  });
+};
+
 export const ScheduledEventsTableRow: FunctionComponent<ScheduledEventsTableRowProps> = (props) => {
+  const [isLoading, setLoading] = useState(false);
+  const [scheduleId, setScheduleId] = useState<number>();
+
+  useEffect(() => {
+    setScheduleId(props.uid);
```
edwinmp: Why not just do this on line 46?

edwinmp: Also, why is `scheduleId` in state? ... I don't see it changing anywhere else

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,52 @@ export interface ScheduledEventsTableRowProps {
...
+
+  const handleClick = () => {
+    setLoading(true);
+    createRunInstance(scheduleId, {
+      start_at: moment(),
+      status: 'p',
+    })
+      .then(() => {
+        setLoading(false);
+      })
```
edwinmp: One line.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,52 @@ export interface ScheduledEventsTableRowProps {
...
+      .catch(() => {
+        setLoading(false);
+      });
```
edwinmp: I'd rather you add a `console.log` of whatever went wrong.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -34,6 +84,11 @@ export const ScheduledEventsTableRow: FunctionComponent<ScheduledEventsTableRowP
           ? `Every ${props.interval} ${props.interval_type}`
           : 'None'}
       </td>
+      <td>
+        <Button variant="outline-danger" size="sm" onClick={handleClick} data-id={props.id}>
+          {isLoading ? 'Loading…' : 'Run'}
```
edwinmp: `Run Now` ...

edwinmp: Seeing where `isLoading` is being used, I don't think that's the right name ... `isRunning` seems more appropriate. Or do we go for `Creating instance` :)

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -34,6 +84,11 @@ export const ScheduledEventsTableRow: FunctionComponent<ScheduledEventsTableRowP
           ? `Every ${props.interval} ${props.interval_type}`
           : 'None'}
       </td>
+      <td>
+        <Button variant="outline-danger" size="sm" onClick={handleClick} data-id={props.id}>
```
edwinmp: Add a class `disabled` for when `isLoading` is true ... that way it won't be clicked again.

edwinmp (changes requested):

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
davidebukali: Removed the blank line, it breaks without `return` :)

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Updated

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Also updated

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Removed unnecessary state

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Updated

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Now logs with a JSON string

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: They are both good options, I think `Creating instance` is better :)

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Updated

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
edwinmp: Hehe ... ofc the structure has to change without the return ... not required though, so never mind....

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,46 @@ export interface ScheduledEventsTableRowProps {
...
+  start_at: Moment;
+  status: string;
+}
+
+const BASEPATH = api.routes.VIEW_SCHEDULED_EVENTS;
+
+const createRunInstance = async (
+  scheduleId: number | undefined,
+  data: RunInstancePayload,
+): Promise<boolean> => {
```
edwinmp: I highly doubt that this returns a boolean on success

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,46 @@ export interface ScheduledEventsTableRowProps {
...
 
+interface RunInstancePayload {
+  start_at: Moment;
+  status: string;
+}
+
+const BASEPATH = api.routes.VIEW_SCHEDULED_EVENTS;
+
+const createRunInstance = async (
+  scheduleId: number | undefined,
```
edwinmp: This can be shorter ... and if you're getting complaints about not being able to have an undefined filed before an always defined one, change their positions.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,46 @@ export interface ScheduledEventsTableRowProps {
...
+  data: RunInstancePayload,
+): Promise<boolean> => {
+  const token = await localForage.getItem<string>(localForageKeys.API_KEY);
+
+  return await axios.request({
+    url: `${BASEPATH}${scheduleId}${api.routes.CREATE_SCHEDULED_INSTANCE}`,
```
edwinmp: `scheduleId` can be undefined, but you're continuing to use it without safeguards ... either handle the undefined scenario, or set it to always be defined... I'm for the latter. It should required by this function.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,46 @@ export interface ScheduledEventsTableRowProps {
...
 export const ScheduledEventsTableRow: FunctionComponent<ScheduledEventsTableRowProps> = (props) => {
+  const [isCreatingInstance, setLoading] = useState(false);
+
+  const handleClick = () => {
+    setLoading(true);
+    createRunInstance(props.uid, {
+      start_at: moment(),
+      status: 'p',
+    })
+      .then(() => setLoading(false))
```
edwinmp: You're assuming coming here always means a successful result ... The API will return a successful event even when authentication fails, so u need to handle those situations here by examining status codes.

frontend/src/utils/api.ts
```diff
@@ -9,5 +9,6 @@ export const api = {
     EXPORT: '/api/export/',
     CHANGE_PASSWORD: '/api/change_password/',
     VIEW_SCHEDULED_EVENTS: '/api/scheduled_event/',
+    CREATE_SCHEDULED_INSTANCE: '/run_instances/',
```
edwinmp: I'm certain this isn't the path ... use the full path for this endpoint ... it's confusing to have it depend on another endpoint & complicates changes.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -12,7 +17,46 @@ export interface ScheduledEventsTableRowProps {
...
+      Authorization: `token ${token}`,
+    },
+    data,
+  });
+};
+
 export const ScheduledEventsTableRow: FunctionComponent<ScheduledEventsTableRowProps> = (props) => {
+  const [isCreatingInstance, setLoading] = useState(false);
+
+  const handleClick = () => {
```
edwinmp: Give this a return type.

edwinmp (changes requested):
This is dependent on @mariamiah's PR #188 ... can't have it merged in before it since a success here should register in that run instances table.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Updated

frontend/src/utils/api.ts
davidebukali: Updated with a template string that takes in the `scheduleId`

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Now check for `200` and `201` status codes, otherwise we log the response

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: `scheduleId` is now required without the `undefined`

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Now only accepts `number` removed `undefined` type; Great idea to make undefined appear before the actual type.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
davidebukali: Updated to reflect json response of the success function.

akmiller01 (dismissed):
Oops, looking at the wrong PR

edwinmp (approved):

PR #191, opened 2020-06-04
Feature/view running instances

### LINKED ISSUE #178: View the run instances of each scheduled event
Dependent on #163
Behaviour here #167

mariamiah:


frontend/src/components/ScheduledEventsRunHistoryTableCard/ScheduledEventsRunHistoryTableCard.tsx
```diff
@@ -0,0 +1,69 @@
...
+        pageCount={pageCount}
+        onPageChange={handlePageChange}
+      />
+    );
+  };
+
+  return (
+    <div className="col-md-12">
+      {historyData && historyData.length > 0 ? (
+        <React.Fragment>
```
edwinmp: Why the fragment?

frontend/src/components/ScheduledEventsRunHistoryTableCard/ScheduledEventsRunHistoryTableCard.tsx
```diff
@@ -0,0 +1,69 @@
...
+        count={count}
+        pageCount={pageCount}
+        onPageChange={handlePageChange}
+      />
+    );
+  };
+
+  return (
+    <div className="col-md-12">
+      {historyData && historyData.length > 0 ? (
```
edwinmp: Just having historyData.length is enough since any number that isn't `0` is truthy
So it becomes `historyData && historyData.length ? (...`

frontend/src/components/ScheduledEventsRunHistoryTableCard/ScheduledEventsRunHistoryTableCard.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <React.Fragment>
+          <Card className="col-md-12">
+            <Card.Header className="card-header-rose card-header-icon">
+              <h4 className="card-title">Update FTS Run History</h4>
```
edwinmp: What if the row I clicked on has nothing to do with FTS?

frontend/src/components/ScheduledEventsRunHistoryTableCard/utils/index.ts
```diff
@@ -0,0 +1,24 @@
+import * as localForage from 'localforage';
+import { api, localForageKeys } from '../../../utils';
+import axios from 'axios';
```
edwinmp: Organise imports

frontend/src/components/ScheduledEventsRunHistoryTableCard/utils/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const BASEPATH = api.routes.VIEW_SCHEDULED_EVENTS;
```
edwinmp: I don't think its a problem how you guys are using this ... these endpoints could change, and then you'd find yourself with lots of broken features ... So for each endpoint, I'm gonna need a separate constant, not one concatenated from another.
i.e. have one for `FETCH_RUN_INSTANCES` ... if you're doing this somewhere else, please fix ... I saw this in David's code as well & think it originated here.

frontend/src/components/ScheduledEventsRunHistoryTableCard/utils/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const fetchRunHistory = async (eventId: number): Promise<any> => {
+  const token = await localForage.getItem<string>(localForageKeys.API_KEY);
+  const headers = {
+    'Content-Type': 'application/json',
+    Authorization: `token ${token}`,
+  };
+
+  return await axios(`${BASEPATH}${eventId}/run_instances/`, { headers });
```
edwinmp: Use a constant with a placeholder that your can then replace with string.replace

frontend/src/components/ScheduledEventsRunHistoryTableCard/utils/index.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export interface RunHistory {
+  status: string;
+  start_at: string;
+  ended_at: string;
+}
+export interface HistoryData {
+  data?: object;
```
edwinmp: Give it a proper type, not object.

frontend/src/components/ScheduledEventsRunHistoryTableRow/ScheduledEventsRunHistoryTableRow.tsx
```diff
@@ -0,0 +1,25 @@
...
+import moment from 'moment';
+import React, { FunctionComponent } from 'react';
+import { convertStatus, getStatusClasses } from './utils';
+
+interface ScheduledEventsRunHistoryTableRowProps {
+  status: string;
+  started: string;
+  ended: string;
+}
+export const ScheduledEventsRunHistoryTableRow: FunctionComponent<ScheduledEventsRunHistoryTableRowProps> = (
```
edwinmp, line 11: Blank line before this...

frontend/src/components/ScheduledEventsRunHistoryTableRow/ScheduledEventsRunHistoryTableRow.tsx
```diff
@@ -0,0 +1,25 @@
...
+  props,
+) => {
+  const badgeClass = getStatusClasses(props.status);
+
+  return (
+    <tr>
+      <td>
+        <span className={classNames('badge', badgeClass)}>{convertStatus(props.status)}</span>
```
edwinmp: Surprised you converted this ... I remember u left the original scheduled events table with its `wks` `mnths` stuff ... not complaining, I prefer this.

frontend/src/components/ScheduledEventsRunHistoryTableRow/utils/index.ts
```diff
@@ -0,0 +1,28 @@
...
+};
+
+const statusClasses: { [char: string]: string } = {
+  p: 'badge-secondary',
+  r: 'badge-warning',
+  c: 'badge-success',
+  e: 'badge-danger',
+};
+export const convertStatus = (statusValue: string): string | undefined => {
+  for (const key in status) {
```
edwinmp: Do you have a system on when to have a blank line or is it just random? I prefer consistency ...

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -1,15 +1,28 @@
...
+  const [activeRow, setActiveRow] = useState(0);
+  const history = useHistory();
+
+  const handleClick = (id: number) => (): void => {
+    if (id !== activeRow) {
+      setActiveRow(id);
+      history.push({
+        pathname: '/scheduledevents/',
+        state: { rowId: id },
+      });
```
edwinmp: Ah, so that's what the history was being used for ... this won't do ... find another way. I don't know what shape the ScheduledEvents API returns them, but it should be providing an ID ... if it's not, please make sure it is. It's that ID that you need to capture on click.

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -24,6 +37,8 @@ export const ScheduledEventsTable: FunctionComponent<ScheduledEventTableProps> =
           interval_type={event.interval_type}
           repeat={event.repeat}
           start_date={event.start_date}
+          onClick={handleClick(offset + index + 1)}
```
edwinmp: This won't do ... Say you have scheduled events 1, 2, and 3 ... What happens when I delete scheduled event 2? 3 will move into position 2, but that ID will no longer exist. Use the individual IDs for each event, those returned by the API.

frontend/src/utils/localForage.ts
```diff
@@ -3,6 +3,7 @@ import * as localForage from 'localforage';
 export const localForageKeys = {
   API_KEY: 'API_KEY',
   USER: 'USER',
+  EVENT_ID: 'EVENT_ID',
```
edwinmp: This is leading to all kinds of weird errors ... you only want to add this to local storage that you need to persist after refresh, and they have to definitely carry some decent significance across multiple pages ... this doesn't qualify. For one, it results in a weird bug where when I reload the page, there's still a run instance showing without a selected event

image

edwinmp (changes requested):
Good work ... lots of creative solutions.

The toggle buttons under 'Enabled' are supposed to be read only ... but now they can be toggled on click... also the two tables are not exactly aligned ...

frontend/src/utils/api.ts
```diff
@@ -9,5 +9,6 @@ export const api = {
     EXPORT: '/api/export/',
     CHANGE_PASSWORD: '/api/change_password/',
     VIEW_SCHEDULED_EVENTS: '/api/scheduled_event/',
+    FETCH_RUN_INSTANCES: '/api/scheduled_event/id/run_instances/',
```
edwinmp: This is risky ... template strings must take a risk free format ... e.g `{ID}` or `%ID%`, something you won't usually find in a random string. Now simply having it as `id` ... imagine the url was `/api/cupid/id/smells/stupid` ... doing a string.replace on that aiming to replace `id` with `MEH` would lead to `/api/cupMEH/MEH/smells/stupMEH`

frontend/src/components/ScheduledEventsRunHistoryTableCard/utils/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  return await axios(`${BASEPATH.replace('id', eventId.toString())}`, { headers });
+};
+
+export const getScheduledEventsByPage = (
+  currentPage: number,
+  data: ScheduledEventRunHistory[],
+): ScheduledEventRunHistory[] => {
+  const begin = (currentPage - 1) * 5;
+  const end = begin + 5;
```
edwinmp: We call these magic numbers ... I'm guessing 5 refers to the limit? What happens when the limit changes? You'll have to update several lines ... so either pass in this value as a parameter, or use the constant... I've seen it somewhere already.

frontend/src/components/ScheduledEventsTableCard/ScheduledEventsTableCard.tsx
```diff
@@ -1,14 +1,18 @@
 import React, { FunctionComponent, ReactNode, useEffect, useState } from 'react';
 import { Card } from 'react-bootstrap';
 import { Dimmer, Loader } from 'semantic-ui-react';
+import { fetchData, getScheduledEventsByPage } from '.';
 import { PaginationRow } from '../PaginationRow';
 import { ScheduledEventsTable } from '../ScheduledEventsTable';
-import { fetchData } from '.';
 
 const LIMIT = 5;
```
edwinmp: You're creating this in too many places ... Have it in a single util file that you use throughout all related code ... as long as its tied to the same table. i.e. ScheduledEvents limit = 5, runhistory limit = 10

frontend/src/components/ScheduledEventsTableCard/utils/index.ts
```diff
@@ -1,9 +1,12 @@
...
 import * as localForage from 'localforage';
+import { ScheduledEvent } from '../../../types/scheduledEvents';
 import { api, localForageKeys } from '../../../utils';
-import axios from 'axios';
 
 const BASEPATH = api.routes.VIEW_SCHEDULED_EVENTS;
 
+export const LIMIT = 5;
+
 export const fetchData = async (): Promise<any> => {
```
edwinmp: No unnecessary `any`

frontend/src/components/ScheduledEventsRunHistoryTableCard/utils/index.ts
```diff
@@ -0,0 +1,26 @@
+import axios from 'axios';
+import * as localForage from 'localforage';
+import { ScheduledEventRunHistory } from '../../../types/scheduledEvents';
+import { api, localForageKeys } from '../../../utils';
+
+const BASEPATH = api.routes.FETCH_RUN_INSTANCES;
+
+export const fetchRunHistory = async (eventId: number): Promise<any> => {
```
edwinmp: No unnecessary `any` ... avoid it like the plague and you'll master typescript types.

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -1,29 +1,39 @@
...
-          id={offset + index + 1}
+          key={event.id}
+          id={event.id}
           name={event.name}
           description={event.description || ''}
           enabled={event.enabled}
           interval={event.interval}
           interval_type={event.interval_type}
           repeat={event.repeat}
           start_date={event.start_date}
```
edwinmp: I probably missed this in the previous scheduled events PR review ... mentioned in for the history though ... instead of passing each property as its own prop, please pass in the entire event. A lot cleaner.

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -1,29 +1,39 @@
...
   currentPage: number;
   pageLimit: number;
   events: Array<{}>;
+  handleRunHistory: (id: number, name: string) => void;
 }
 export const ScheduledEventsTable: FunctionComponent<ScheduledEventTableProps> = (props) => {
-  const offset = (props.currentPage - 1) * props.pageLimit;
+  const [activeRow, setActiveRow] = useState(0);
+
+  const handleClick = (id: number, name: string) => (): void => {
```
edwinmp: Glad to see you do this ... shows a lot better understanding of JS... Anyway, make this simpler ... just a function that takes in the event (or id & name), but shouldn't return a function... all that event info should be passed in from the `ScheduledEventsTableRow` component.

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -1,29 +1,39 @@
...
+          key={event.id}
+          id={event.id}
           name={event.name}
           description={event.description || ''}
           enabled={event.enabled}
           interval={event.interval}
           interval_type={event.interval_type}
           repeat={event.repeat}
           start_date={event.start_date}
+          onClick={handleClick(event.id, event.name)}
```
edwinmp: Should just be `onClick={handleClick}`

frontend/src/components/ScheduledEventsTableCard/utils/index.ts
```diff
@@ -13,3 +16,13 @@ export const fetchData = async (): Promise<any> => {
 
   return await axios(`${BASEPATH}`, { headers });
 };
+
+export const getScheduledEventsByPage = (
+  currentPage: number,
+  data: ScheduledEvent[],
+): ScheduledEvent[] => {
+  const begin = (currentPage - 1) * 5;
+  const end = begin + 5;
```
edwinmp: Replace these magic numbers with constant or parameter.

frontend/src/components/ScheduledEventsTableRow/ScheduledEventsTableRow.tsx
```diff
@@ -10,18 +11,20 @@ export interface ScheduledEventsTableRowProps {
   repeat: string;
   interval_type: string;
   start_date: string;
+  onClick?: () => void;
```
edwinmp: This one should take in the event, or id & name ... i.e. `onClick?: (event: ScheduledEvent) => void;`

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -1,29 +1,39 @@
...
+import React, { FunctionComponent, ReactNode, useState } from 'react';
 import { Table } from 'react-bootstrap';
 import { ScheduledEvent } from '../../types/scheduledEvents';
 import { ScheduledEventsTableRow } from '../ScheduledEventsTableRow';
-
 export interface ScheduledEventTableProps {
   currentPage: number;
   pageLimit: number;
   events: Array<{}>;
+  handleRunHistory: (id: number, name: string) => void;
```
edwinmp: Rename this to `onRowClick`, or `onRowSelect` ... it's not likely to be just about the run history ...may want to do something else when a user selects a row. Rename for all other occurrences, even for `ScheduledEventsTableCard`

frontend/src/components/ScheduledEventsTable/ScheduledEventsTable.tsx
```diff
@@ -1,29 +1,39 @@
...
+import classNames from 'classnames';
+import React, { FunctionComponent, ReactNode, useState } from 'react';
 import { Table } from 'react-bootstrap';
 import { ScheduledEvent } from '../../types/scheduledEvents';
 import { ScheduledEventsTableRow } from '../ScheduledEventsTableRow';
-
 export interface ScheduledEventTableProps {
   currentPage: number;
   pageLimit: number;
   events: Array<{}>;
```
edwinmp: I'm sure these can have a proper type.

edwinmp (changes requested):

edwinmp (approved):
LGTM!

PR #197, opened 2020-06-22
Feature/fts update api

wakibi:
@akmiller01 kindly review this


akmiller01:
@wakibi is it possible to rebase your branch with the changes from develop? This still has your commits changing the .R "plyr" reference and the .csvs which no longer need to be changed.

Also, I like your pypika delete and insert into logic, but I am concerned about coding this singularly around FTS. Especially when we have generic script executing logic. I would prefer if we could have a generic table update endpoint and a generic table download endpoint.

Thoughts @edwinmp @davidebukali @mariamiah ?

wakibi:
@akmiller01 I believe it can be easily extended / changed to add anything else. I just was not going to open up all tables, hence the need for specific tables. Otherwise more tables can always be added. As for the function / class names, that's just a renaming issue and we are good. Ditto the URLs. Let me rebase and push an update to this branch.

edwinmp:
@wakibi could you add some basic documentation to this. Since the endpoints are generic, I need a reference to the shape of the data they consume or produce, especially since I'm going to hardcode each updatable table on the FE & validate the data before pushing it.

edwinmp (approved):
I've merged in Alex's changes ... if there's nothing else holding this back, could please merge it in @wakibi

PR #198, opened 2020-06-22
Linting and removal of FTS specific API Views

akmiller01:


akmiller01:
Hey @wakibi here are my suggested edits for your feature branch. Given the execute_update API endpoint we don't need the specific script runners for FTS and the related utilities. Posted a screenshot in Slack of it working for fts_diff.sh for e.g.

Also I went ahead and removed the authentication from your streaming all-data endpoint, since this is limited by table it would be nice to open this endpoint up for API users without the need to authenticate.

edwinmp (approved):

PR #204, opened 2020-06-30
Start on IATI Registry work

akmiller01:


wakibi (approved):
So far run through 2000 of the 8000+ xml files, looks good. Approving in the meantime as I run tests on the rest.

PR #219, opened 2020-07-07
Feature | ddw models

edwinmp:
Was having trouble knowing what we had in the warehouse, so I added the missing models.
These should count as dev metadata, and we should add to them as we go - some descriptions, maybe even add the aliases Bill was talking.


akmiller01:
Hey @edwinmp I was actually thinking about this recently.

Technically speaking, the table definitions are all held in `data_updates/manual/CSV/meta.csv` and `data_updates/manual/CSV/meta_columns.csv` but it's a bit of a clunky solution.

Django models are similarly limited, in that we would need to re-write them every time there is a change to the data structure.

Do you think it would be possible to abstract table models from the database automatically, if not on the fly?

edwinmp:
> Django models are similarly limited, in that we would need to re-write them every time there is a change to the data structure.
> 
> Do you think it would be possible to abstract table models from the database automatically, if not on the fly?

Not sure how often or how extensively the data structure is likely to change, but these were auto generated quite easily, just needed a clean up. Not sure it'd be wise to automate that. The only way I can think of to make this one way is to manage the tables exclusives within Django.

wakibi:
> > Django models are similarly limited, in that we would need to re-write them every time there is a change to the data structure.
> > Do you think it would be possible to abstract table models from the database automatically, if not on the fly?
> 
> Not sure how often or how extensively the data structure is likely to change, but these were auto generated quite easily, just needed a clean up. Not sure it'd be wise to automate that. The only way I can think of to make this one way is to manage the tables exclusives within Django.

I think we can do with both worlds, perhaps use one to check the other. If it's for the sole purpose of knowing what the DB schema looks like for each table, then this must be easier.

akmiller01 (approved):

Issue #217, opened 2020-07-07
Cannot create run instance via Run Now button when there's one that's pending


edwinmp:
The idea was to avoid creating a lot of closely grouped instances, but currently, an instance cannot be created even if the pending one is set to run in the distance future.

I suggest we include a time element to the logic ... i.e a new runtime instance cannot be created when there's an existing one that's either running, or set to run within the next 5 mins. This must be accompanied by visual feedback, an alert explaining why the instance wasn't created.

PR #222, opened 2020-07-08
fix/create_multiple_run_instances

davidebukali:
#217


core/serializers.py
```diff
@@ -281,11 +285,27 @@ class Meta:
             'status'
         )
 
+    def isInstanceRunning(self, scheduled_event):
```
edwinmp: It's python ... they do snake case for functions and variables.

core/serializers.py
```diff
@@ -281,11 +285,27 @@ class Meta:
...
+        queryset = ScheduledEventRunInstance.objects.filter(
+            (Q(scheduled_event=scheduled_event.id) & Q(status='r'))
+        )
+        if queryset.exists():
+            return queryset.earliest('id')
+
+    def isDueToRunIn5Minutes(self, scheduled_event):
```
edwinmp: Same here.

core/views.py
```diff
@@ -418,12 +419,35 @@ def get(self, request, pk, format=None):
             serializer = ScheduledEventRunInstanceSerializer(scheduled_event_run_instance, many=True)
             return Response(serializer.data)
 
+    def getPostResponse(self, serializer, request):
```
edwinmp: And here...

edwinmp (changes requested):

core/serializers.py
davidebukali: Updated

core/serializers.py
davidebukali: Updated

core/views.py
davidebukali: Updated

edwinmp (approved):

Issue #249, opened 2020-07-27
Boolean fields don't display on the front end


akmiller01:
I've imported the column "x_covid" from the "iati_transactions" table as a Boolean. You're able to filter the column on the front end by selecting "is equal" and putting the strings "TRUE" or "FALSE" and they appear when exporting the CSV, but the boolean values print as blank strings on the preview table:

![image](https://user-images.githubusercontent.com/2836840/88563071-81d43680-cfff-11ea-8860-f22dea705669.png)

Worth coding in a display for boolean fields @edwinmp ? Or do we want to work-around this for now and just made "x_covid" a string field?

edwinmp:
Since the API returns proper booleans, this should be a proper FE fix ... should be simple.

Issue #237, opened 2020-07-23
Confirmation before deleting a query


bill-anderson:
Add a confirmation message before deleting a query.

(I have just deleted a whole query, thinking I was only deleting the step closest to the delete button)

edwinmp:
@bill-anderson the delete buttons are a lot clearer now so you'll know what you're deleting, and they come with a confirmation. First click changes the button caption from `Delete` to `Confirm Delete` ... a second click deletes the item.
![image](https://user-images.githubusercontent.com/5672438/88808280-58f8a080-d1bb-11ea-9c76-d9c36c644efc.png)

PR #250, opened 2020-07-27
Feature | confirm delete

edwinmp:
Fix for #237 #249 

You're now required to click delete buttons twice to delete. The first time asks for a confirmation (button caption changes from `Delete X` to `Confirm Delete X`. Click again to delete. If the confirmation doesn't come in 3 secs, it resets.

@akmiller01 added the boolean fix here as well.


akmiller01 (approved):
Looks good. I'll let you know when IATI is done processing and then we can go live? I had to reset the script like 4 times over the weekend.

PR #271, opened 2020-07-29
Fix | scheduled events

edwinmp:
Added logs to the run instances and cleaned up the management command logic a bit
image

It's likely the live scripts are failing due to a permissions issue - it's the most common one I ran into testing locally.


akmiller01 (approved):
:+1:

PR #272, opened 2020-07-30
Remove authentication for csv export api

davidebukali:
@akmiller01 this is from the conversation on slack about authentication free `csv` exports


akmiller01 (approved):
Looks good!

PR #267, opened 2020-07-29
Feature/preview dataset results

davidebukali:
#243

Issue #235, opened 2020-07-23
Limit size of export csv (no. of rows)


bill-anderson:
If a query will return more than 1,048,575 rows (excel limit) it should be abandoned and the user should receive a message to rethink their query.

k8hughes:
This one sounds like it would be useful to have especially now we are working with IATI data.  Moved to `to do`.

edwinmp:
@davidebukali you seem to have worked on this at some point ... what happened?

k8hughes:
@davidebukali Can you update this issue to say what work was done on this in the end?

davidebukali:
@k8hughes @edwinmp the previous PR was closed because it could not scale for larger datasets (counting rows for each query and comparing with the limit). Adding a flag to the query to limit rows before they are returned is a pending fix.

PR #283, opened 2020-08-06
Feature/limit export csv

davidebukali:
Feature for issue #235 It depends on `PreviewOperationData` api endpoint from #267


edwinmp:
Are you sure this can be done in a performant way? Counting a million rows is definitely super slow ... I'd rather you just add a MAX_LIMIT and leave it at that.

PR #314, opened 2020-09-07
Feature | integrate celery

edwinmp:
The `post_save` that I'd setup to count rows was slowing down saving of datasets... doing it asynchronously via celery solves that.


akmiller01 (approved):
Looks good!

Issue #316, opened 2020-09-08
Send email notification when an scheduled event fails or runs for too long


edwinmp:
Use the Django [ADMINS](https://docs.djangoproject.com/en/3.1/ref/settings/#admins) setting to determine who receives these emails

k8hughes:
has this been deployed?  I can see in the thread that something has been merged.  If functional review is required can someone be tagged in this issue to do that for us if not then lets close it 👍

PR #326, opened 2020-09-11
Feature/email notifications

davidebukali:
Send email notification to users in 'ADMINS=[]' setting variable when a scheduled event has run for too long or failed; More details #316


akmiller01:
Hi @davidebukali , just testing this out now.

Do I need to fill out these alert fields in order for this to work?
image

It might be easier to have the alert messages be programmatically determined from the scheduled event name, and the type of error that happened.

davidebukali:
@akmiller01 I agree its easier to create the message programmatically, maybe scrap the `Alert` class and send mail from the `ScheduledEvent` class; what do you think ?

akmiller01:
> @akmiller01 I agree its easier to create the message programmatically, maybe scrap the `Alert` class and send mail from the `ScheduledEvent` class; what do you think ?

Yes that sounds like a good way of handling it!

davidebukali:
@akmiller01 emails are now created programmatically from the `scheduled_event` using `script_name` and error that occured

akmiller01 (approved):
Looks good, David!

Issue #254, opened 2020-07-28
Make query description mandatory


bill-anderson:
The point of publishing a query is for others to use it. For this it needs a clear description. So if 'Draft' is not checked description should be mandatory.

PR #327, opened 2020-09-17
Make query description mandatory for published datasets

davidebukali:
If dataset is a draft, description should be optional and required if its not a draft(published) #254


akmiller01:
Looks good to me, but I'm completely unfamiliar with the typescript files, so also requesting review from @edwinmp !

edwinmp (approved):
LGTM!

Issue #258, opened 2020-07-28
Change datasets table UI to show more information about the dataset


edwinmp:
Currently, we have a table. It's not the most extensible choice. A custom list view would be better.
![image](https://user-images.githubusercontent.com/5672438/88650460-6df90500-d0d1-11ea-8a96-4fbdfbd84c3c.png)

Would like to see the following:
- [x] Name
- [x] Description
- [x] Created By
- [x] Updated On
- [x] Status (a badge)
- [x] Data Source
- [x] Collapsible view of the generated SQL query
- [ ] Number of steps (optional)
- [x] Action Buttons (Edit, View Data)

edwinmp:
Updated design

![image](https://user-images.githubusercontent.com/5672438/92690229-4ee4c880-f349-11ea-8c65-7dbfa224a151.png)

edwinmp:
Design local URL http://127.0.0.1:8000/designs/datasets/

edwinmp:
@bill-anderson this has been deployed.

bill-anderson:
Thanks, looks good.
 
But ...
When I save a query I get this error. The query has in fact saved ...

![ddw-save](https://user-images.githubusercontent.com/3491562/95112715-380f7700-0739-11eb-82b8-cccb4fe14b81.png)

edwinmp:
@bill-anderson that resulted from a error during deployment ... it's fixed.

bill-anderson:
Thanks
Great job.
One last thing. Can we included my datasets that have been published (not drafts) in Published datasets?

edwinmp:
Sure ... will do.

edwinmp:
> Can we included my datasets that have been published (not drafts) in Published datasets?

@bill-anderson this is done.

PR #308, opened 2020-08-27
Update datasets UI

mariamiah:
Issues #258

PR #341, opened 2020-09-23
Feature/my datasets ui

davidebukali:
Update datasets table ui to use a listview #258


davidebukali:
This was meant to have code for `MyDatasets` page; closing now because it has a lot of similar code to `PublishedDatasets`. All this work now pushed and handled at #308

PR #344, opened 2020-09-30
Feature/freeze unfreeze api

wakibi:
Please make sure the two schemas are created first i.e `CREATE SCHEMA archives;` and `CREATE SCHEMA dataset;` before testing


README.md
```diff
@@ -109,7 +109,14 @@ To create a test development DB, for local development (e.g. virtualenv steps be
...
 
-### Scheduled Events
+### Freeze/Unfreeze and Save Queryset features
+Make sure the following schemas are created;
+
+- `archives` i.e `CREATE SCHEMA archives;`
+- `dataset` i.e `CREATE SCHEMA dataset;`
+
+### Django-crontab setup
+This is used to run cron jobs that run automated scripts that are added from the UI. You need to make sure cron is installed on the web docker container. If not, run `apt-get install cron`. This has been added to the Dockerfile so it should install this automatically for new deployments (container rebuilds)
```
edwinmp: I hope this doesn't interfere with native crons, specifically those that run the scheduled events ... we removed this cos it resulted in context errors

core/docs/API.md
```diff
@@ -61,3 +61,21 @@
...
+    Get - https://ddw.devinit.org/api/savedquerysets/
+
+13. Return details of Saved Query Data selected by ID
+    Get - https://ddw.devinit.org/api/savedqueryset/[SAVEDQUERYDATA_ID]/
+
+14. Download Frozen Data
+    Get - https://localhost/api/tables/download/[FROZEN_TABLE_NAME]/archives/
+
+15. Download Saved Query Set Data
+    Get - https://localhost/api/tables/download/[QUERY_SET_DB_TABLE]/dataset/
```
edwinmp: Given that the initial context is the data source e.g. FTS ... how do I fetch only frozen datasets for a particular datasource?

Also 14 & 15 are linking to localhost ... ah, and the numbering is wrong.

core/models.py
```diff
@@ -95,8 +97,8 @@ class Operation(BaseEntity):
     sample_output_path = models.TextField(blank=True, null=True)
     tags = models.ManyToManyField(Tag)
     is_draft = models.BooleanField(default=True)
-    row_count = models.IntegerField(blank=True, null=True)
-    count_rows = models.BooleanField(default=False) # controls whether to count rows in the post_save signal
```
edwinmp: The reason these were removed? This is going to break more than a few things

core/pypika_utils.py
```diff
@@ -77,7 +79,8 @@ def __init__(self, operation=None, operation_steps=None, source=None):
             self.initial_table_name = query_steps.first().source.active_mirror_name
             self.initial_schema_name = query_steps.first().source.schema
 
-        self.current_dataset = Table(self.initial_table_name, schema=self.initial_schema_name)
+        self.current_dataset = Table(
+            self.initial_table_name, schema=self.initial_schema_name)
```
edwinmp, line 83: You have a lot of these formatting changes across multiple files ... curious, are u doing them manually, or is it some automated linter doing it?

edwinmp: If it's a linter, let me know which one it is ... best we agree on these things to avoid clashes, cos if we all use different ones, we're likely to keep pushing changes that don't add new functionality but simply undo each other's code styles ...

core/serializers.py
```diff
@@ -55,6 +57,14 @@ def to_representation(self, instance):
                 ]
             }
 
+    def estimate_count(self):
+        if self.operation:
+            MAX_OFFSET = 65390
+            count, data = query.query_table(
+                self.operation, 1, MAX_OFFSET, estimate_count=True)
+            return len(data) > 0
+        return True
```
edwinmp: I found this to be wildly inaccurate ... which is why I added the asynchronous `row_count` feature that returns an accurate count via celery ... actually, this looks like something I wrote a long time ago that I have since removed ... I worry you've got a bit of old code.

edwinmp (changes requested):
80% of the changes are formatting, plus there's some signs of old code that makes me worry about rebase-based undoing of deployed code. This can be avoided by rebasing often & carefully each time the `develop` branch is updated. I suggest u go through this `File changes` section and make note of changes u made & those that are a result of botched conflict resolution during a rebase.

wakibi:
Please hold as I fix the linting...

core/pypika_utils.py
wakibi: It's a linter, `pylint`.

README.md
wakibi: This has been removed. Rebasing issue.

core/serializers.py
wakibi: Removed

core/models.py
wakibi: Removed

core/docs/API.md
wakibi: > Given that the initial context is the data source e.g. FTS ... how do I fetch only frozen datasets for a particular datasource?
> 

I worked on the assumption that you download a dataset one at a time, they can grow very huge! You will need the `frozen table name` and one of two schemas i.e dataset for saved querysets or archives for frozen data

> Also 14 & 15 are linking to localhost ... ah, and the numbering is wrong.

Localhost and numbering fixed.

wakibi:
Using #347 instead.

PR #347, opened 2020-10-01
Feature/freeze unfreeze api migrations

akmiller01:


edwinmp:
@akmiller01 @wakibi Are we keeping this one instead of #344? If so, we should close #344 ...

edwinmp:
@wakibi how exactly does this work? Not clear on how to use this on the FE...

core/models.py
```diff
@@ -254,3 +257,34 @@ class ScheduledEventRunInstance(BaseEntity):
...
+    status_choices = (
+        ('p', 'Pending'),
+        ('r', 'Running'),
+        ('c', 'Completed'),
+        ('e', 'Errored'),
+    )
+
+    parent_db_table = models.CharField(max_length=200, null=False)
+    frozen_db_table = models.CharField(max_length=200, null=False)
+    completed = models.CharField(max_length=1, choices=status_choices, default='p')
```
edwinmp: May I rename this to `status`? Not sure what's happening in the background, so don't want to just change it ...

core/models.py
```diff
@@ -254,3 +257,34 @@ class ScheduledEventRunInstance(BaseEntity):
...
+        ('p', 'Pending'),
+        ('r', 'Running'),
+        ('c', 'Completed'),
+        ('e', 'Errored'),
+    )
+    saved_query_db_table = models.CharField(max_length=200, null=True)
+    active = models.BooleanField(default=True)
+    operation = models.ForeignKey(Operation, on_delete=models.CASCADE)
+    full_query = models.TextField(null=False)
+    completed = models.CharField(max_length=1, choices=status_choices, default='p')
```
edwinmp: This too...

wakibi: Yes please, you can rename them, both. For now they are largely inconsequential and the new name would still make sense. Reason for adding them was in case performance slows down, we can decide to just log the _create requests_ in the DB (to improve user experience and turn around time) and have a cron running in the background to actually finish the job. Was only trying to look into the future.

edwinmp: Yeah, always best to look far ahead ... alright, let me rename them. That's pretty much it 👍

k8hughes:
@edwinmp should be the one to approve this PR as Chris did the original work.  
Will you add your FE work to this before approval?

edwinmp (approved):

PR #351, opened 2020-10-06
Fix | published datasets

edwinmp:
This [comment](https://github.com/devinit/ddw-analyst-ui/issues/258#issuecomment-703805725)


edwinmp:
Need to test this on staging first ... waiting on my laptop that has access to all the servers. Not urgent so don't want to go through the trouble of giving my temporary laptop access

edwinmp:
> Need to test this on staging first ... waiting on my laptop that has access to all the servers. Not urgent so don't want to go through the trouble of giving my temporary laptop access

Done!

PR #355, opened 2020-10-13
Update WEO Oct 2020

akmiller01:


data_updates/R/weo.R
```diff
@@ -25,11 +25,12 @@ table.name = "weo"
 table.quote = c("repo",table.name)
 
 # Load data, removing na strings
-# data_url = "https://www.imf.org/external/pubs/ft/weo/2018/01/weodata/WEOApr2018all.xls"
-data_url = "https://www.imf.org/external/pubs/ft/weo/2019/01/weodata/WEOApr2019all.xls"
-weo = read.csv(data_url,sep="\t",na.strings=c("","n/a","--"))
+data_url = "https://www.imf.org/~/media/Files/Publications/WEO/WEO-Database/2020/02/WEOOct2020all.ashx"
+weo = read.csv(data_url,sep="\t",na.strings=c("","n/a","--"),fileEncoding = "utf-16")
+weo$X = NULL
+weo = subset(weo,WEO.Country.Code!="International Monetary Fund, World Economic Outlook Database, October 2020")
```
wakibi, line 31: Shall we keep editing these lines whenever the URL or encoding changes? I am thinking of passing these as arguments to the R script, or as part of a config file.

data_updates/R/weo.R
```diff
@@ -25,11 +25,12 @@ table.name = "weo"
...
 
-value_cols = paste0("X",c(1980:2024))
+value_cols = paste0("X",c(1980:2025))
```
wakibi, line 33: Same here. Can't we pass these as arguments or configs?

akmiller01: It would be nice to pass them as configs, but it's not as simple as updating the config each year. For e.g. this year they randomly changed the encoding, and added a blank column that gets pulled in as X that would have broken the process if I didn't manually remove it. The main code needs to be tested prior to running, so just making it a configuration wouldn't necessarily save time.

wakibi (approved):

PR #361, opened 2020-10-20
Feature | historical data

edwinmp:
Added a `Versions` button to datasets & data sources
image

image

For either, use the `Freeze Current Version` button to save a new instance
image

Fully saving a new instance is asynchronous, so I added a refresh button. The status should change from `pending` to either `errored` or `completed`. Testing locally will require celery.
`Errored` comes with logs which can be viewed by clicking the `Logs` button
image

Successful instances can be downloaded in the CSV format by clicking the download button.


akmiller01 (approved):
Excited to see this running!

PR #369, opened 2020-11-09
Feature | Text Filtering

davidebukali:
Table columns require a prefix `x_` to be parsed eg `x_recipient_code = 'KE' OR x_recipient_code = 'UG' `

Sample query `x_name = 'Cyclops' AND x_age >= '27' OR x_name = 'Wolverine' OR x_age = '45' OR (x_name = 'Magneto'  AND x_age <= '61')` 

Gives JSON `[[{"field":"name","func":"=","value":"'Cyclops'"}],[{"field":"age","func":">","value":"'27'"}],[{"field":"age","func":"=","value":"'27'"}],{"field":"name","func":"=","value":"'Wolverine'"},{"field":"age","func":"=","value":"'45'"},{"or_brackets":[[{"field":"name","func":"=","value":"'Magneto'"}],[{"field":"age","func":"<","value":"'61'"}],[{"field":"age","func":"=","value":"'61'"}]]}]`

Check console log for output

textareaChange


frontend/src/components/BasicTextarea/BasicTextarea.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const BasicTextarea: FunctionComponent<BasicTextareaProps> = ({ alerts, onChange }) => {
+  const onTextareaChange = (event: ChangeEvent<FormControlElement>) => {
+    onChange(event.currentTarget.value);
+  };
+
+  return (
+    <Form.Group>
+      <Form.Label className="bmd-label-floating">Fill in your query below:</Form.Label>
+      <Form.Control name="description" as="textarea" onChange={onTextareaChange} />
```
edwinmp: If you're going to build a generic component, then everything in it must be generic ... `BasicTextarea` makes it sound like I can use this anywhere I want a text area ... not the case since it has a fixed label and name.

edwinmp: Look through some of the other feedback before u process this one though ... have a feeling u won't need it.

frontend/src/components/OperationStepForm/OperationStepForm.tsx
```diff
@@ -37,6 +37,7 @@ const schema = Yup.object().shape({
 const queries = [
   { key: 'select', icon: 'check', text: 'Select', value: 'select' },
   { key: 'filter', icon: 'filter', text: 'Filter', value: 'filter' },
+  { key: 'advanced', icon: 'rocket', text: 'Advanced Filter', value: 'advanced' },
```
edwinmp: Give it the same icon as the normal filter.

edwinmp: And name it accordingly ... `advanced` could be anything.

frontend/src/components/OperationStepForm/OperationStepForm.tsx
```diff
@@ -89,6 +90,7 @@ export const OperationStepForm: FunctionComponent<OperationStepFormProps> = (pro
 
   const onUpdateOptions = (options: string) => {
     const step = props.step.set('query_kwargs', options);
+    console.log(JSON.stringify(step));
```
edwinmp: ??

frontend/src/components/OperationStepForm/OperationStepForm.tsx
```diff
@@ -97,6 +99,9 @@ export const OperationStepForm: FunctionComponent<OperationStepFormProps> = (pro
     if (query === 'filter') {
       return validateFilter(step);
     }
+    if (query === 'advanced') {
+      return validateAdvanced(step);
```
edwinmp: advanced what?

frontend/src/components/OperationStepForm/OperationStepForm.tsx
```diff
@@ -147,6 +152,28 @@ export const OperationStepForm: FunctionComponent<OperationStepFormProps> = (pro
...
+      setAlerts({
+        error: `These columns donot exist: ${advanced_query['validation']}`,
+      });
+
+      return false;
+    } else if (advanced_query) {
+      return true;
+    } else {
+      setAlerts({
+        error: 'At least one query filter is required, please prefix columns with "x_"',
```
edwinmp, line 168: This `x_` prefix is bound to bite you in the butt ... there's lots of IATI columns that have it.

frontend/src/components/QueryBuilderHandler/QueryBuilderHandler.tsx
```diff
@@ -140,6 +143,17 @@ class QueryBuilderHandler extends React.Component<QueryBuilderHandlerProps> {
...
+    if (query === 'advanced') {
+      const columns = source.get('columns') as ColumnList;
+      const selectableColumns = getStepSelectableColumns(step, steps, columns) as Set<string>;
+
+      const onTextareaChange = (options: string) => {
+        const queryObject = parseAdvancedQueryString(options, selectableColumns);
+        onUpdateOptions(JSON.stringify(queryObject));
+      };
+
+      return <BasicTextarea onChange={onTextareaChange} alerts={alerts} />;
```
edwinmp: Each of the options in this function was neat ... e.g. `SelectQueryBuilder`, `FilterQueryBuilder`, `TransformQueryBuilder` ... then we have this ... why is you crowding our nice code? Basically, create a component to handle all your advanced filter stuff ... I expect it to feed into the `onUpdateOptions` function since it too will return JSON. In this component, I don't want to know how an advanced filter is built.

frontend/src/components/QueryBuilderHandler/utils/index.ts
```diff
@@ -0,0 +1,149 @@
+import Tokenizr from 'tokenizr';
+import { Set } from 'immutable';
+
+export const parseAdvancedQueryString = (options: string, columns: Set<string>): any => {
```
edwinmp: Curious ... can this build anything besides filters? Also will be used outside of Filter components? If not, move it out of the general utils folder for the query builder.

edwinmp: Also, I'm sure you can get rid of all these `any` types.

edwinmp (changes requested):
I think you made a mistake starting with the FE, but too late for that ... you've come up with a config that you feel should work, now it's time to force it to work on the backend.
Remember, be certain that it won't break existing filters.

Also, there's a few things u mentioned in standups that I don't see here ... where's the DraftJS?

frontend/src/components/OperationStepForm/OperationStepForm.tsx
davidebukali: Updated icon and naming

frontend/src/components/BasicTextarea/BasicTextarea.tsx
davidebukali: I found `DraftJS` overkill since both `react-bootstrap` and `semantic-ui` had textareas; 

Updated label to be a `prop`

frontend/src/components/OperationStepForm/OperationStepForm.tsx
davidebukali: Remove console log

frontend/src/components/OperationStepForm/OperationStepForm.tsx
davidebukali: Updated to appropriate naming

frontend/src/components/OperationStepForm/OperationStepForm.tsx
davidebukali: Switched to `col_`

frontend/src/components/QueryBuilderHandler/QueryBuilderHandler.tsx
davidebukali: Created `TextFilterQueryBuilder` to build text filters

frontend/src/components/QueryBuilderHandler/utils/index.ts
davidebukali: Moved `parseAdvancedQueryString` function to appropriate utils folder and updated `any` types

core/pypika_utils.py
```diff
@@ -167,12 +167,140 @@ def filter(self, filters):
         filter_operations = [filter_mapping[filter["func"]](getattr(
             self.current_dataset, filter["field"]), filter["value"]) for filter in filters]
         filter_operations_or = reduce(operator.or_, filter_operations)
+
+        print('filter query string is')
+        print(filter_operations_or)
```
edwinmp: I really should stop seeing these in your code ...

frontend/src/components/BasicTextarea/BasicTextarea.tsx
```diff
@@ -0,0 +1,36 @@
...
+        as="textarea"
+        onChange={onTextareaChange}
+        defaultValue={defaultValue}
+      />
+      <Form.Control.Feedback type="invalid" className="d-block invalid-feedback">
+        {alerts && alerts.error}
+      </Form.Control.Feedback>
+    </Form.Group>
+  );
+};
```
edwinmp: I don't see the point of this component ... inline all this in the one place it's being used.

frontend/src/components/OperationStepForm/OperationStepForm.tsx
```diff
@@ -37,6 +37,7 @@ const schema = Yup.object().shape({
 const queries = [
   { key: 'select', icon: 'check', text: 'Select', value: 'select' },
   { key: 'filter', icon: 'filter', text: 'Filter', value: 'filter' },
+  { key: 'text_filter', icon: 'filter', text: 'Text Filter', value: 'text_filter' },
```
edwinmp: Could you please rename it as `Advanced Filter` ... this make it seem like it's filtering text ... rename in all other occurrences as well.

edwinmp (changes requested):

edwinmp:
@davidebukali Could you please add something to help users out on which columns are available to filter against ... something the shows the mapping of their original names to the expected format. That way they can copy the expected column names and avoid errors ... this is especially useful since you're not doing any validation. See the screenshot below for what happened when I changed `col_` to `co_` and saved the step
image

k8hughes:
@davidebukali could you finish this off so it can be deployed?

core/pypika_utils.py
davidebukali: Removed debugging code

frontend/src/components/BasicTextarea/BasicTextarea.tsx
davidebukali: Updated

frontend/src/components/OperationStepForm/OperationStepForm.tsx
davidebukali: Updated

davidebukali:
> @davidebukali Could you please add something to help users out on which columns are available to filter against ... something the shows the mapping of their original names to the expected format. That way they can copy the expected column names and avoid errors ... this is especially useful since you're not doing any validation. See the screenshot below for what happened when I changed `col_` to `co_` and saved the step
> image

Added modal that maps original column names to the expected format and allows users to copy with a button click.

edwinmp (dismissed):

k8hughes:
This will be added to the advanced query builder to blocked from releasing just now.

Issue #181, opened 2020-06-01
Add an advanced query builder for performing operations between datasets


edwinmp:
_(Pre-requisite - dataset queries indicate which data source they're built on - already exists I think.)_

- Add two options under the `Query Builder` on the sidebar: `Basic` & `Advanced`
- Add the current Query Builder under `Basic`

**Advanced Query Builder**
_Ignore steps below - design has yet to be decided_
- After selecting Active Data Source, show another drop-down for selecting Data Set _(This loads datasets derived from the selected data source)_
- On Save _(Pre-requisite - add a new foreign key field to an operation step for 'dataset`)_
- Lots of testing

PR #377, opened 2020-12-07
Feature | Sub-Query UI Updates

edwinmp:
Issue #181 (to be done as part of the advanced query builder)


k8hughes:
The branch needs cleaning, but need to wait for the FE work.

PR #379, opened 2020-12-10
Feature/estimate query time

wakibi:
Sample return value: [{"result":"success","message":0.54}]
The value of the message float type in micro-seconds for a successful result, or a descriptive error in case an error occurs.


akmiller01 (approved):
Looks good. Working well.

PR #381, opened 2020-12-14
Expose Postgres on port 65432

### LINKED ISSUE #376: Make spotlights data available on the DDW UI
This is what @wakibi wrote on slack:
The data is on the same server as the DDW, but to be able to put that data on the DDW under the current set up, we shall need to do the following;
1. Expose the DDW DB port so the Spotlights API can access it or Set up the spotlights API to run from  a docker container on the same server as the DDW
2. Spotlights data uses 5 schemas (3 active, 2 redundant), we shall need to factor that in, and add them under the sources table in the DDW.
3. Cross check and make sure no table names clash.
One above is the most tricky one as I have never done port forwarding to a docker container before.

@akmiller01 have you ever done port forwarding to a docker container?

### LINKED ISSUE #394: Closing the old Linode server account
Below is the email thread disscussing what is still active on this server.
Please can we make sure that we move what needs moving and close this account by the end of the month?

I'm tagging @akmiller01 @wakibi  and @edwinmp on this issue.  Can we distill this email thread into the clear action point we have to get the task done please?

@wakibi : No, the data is coming off the current Analyst UI server; all spotlights data at least. It’s just the API code that needs to be hosted elsewhere.
 

From: Alex Miller

Hi Chris,

Thanks for taking a look. Since https://api.devinit.org is still running, I presume it's still feeding from linode1009616_ddw? So presumably we need to re-host the API code on a different server, and then reconfigure the API code to point to the Analyst UI server instead of the old DDW before we can shut down the rest of the services.

Best,
-Alex
 
Alex Miller I Data Scientist

 
From: Chris Wakibi
 
Alex,
 
https://api.devinit.org is still running the spotlights API as you mentioned. We need to move it somewhere else if we are to shut down this server. Have looped in @Edwin Magezi as he does the FE for the Spotlights and it could be affected by any migrations (assuming some URLs are hard-coded IP addresses)
 
Also, all DDW data that was on (linode1009616_ddw) was moved to the Analyst UI server. I do not see anything else running off that server.
 
Christopher Wakibi | Senior Fullstack Developer

From: Alex Miller

Hi all,

Here's what remains running on the server:

I believe the Spotlight data was transferred off of the original DDW (linode1009616_ddw), but I don't know if any of the other data was archived. If we can dump the contents of the database and store it somewhere secure, that would be easy to shutdown and accounts for $82/month.

linode3186698_juds is running http://joinedupdata.org/en/about/ and costs $20/month.

linode3673430_dh_staging is still running https://api.devinit.org/ and I thought this had been migrated somewhere else as well but maybe I misunderstood. Do the spotlights still depend on this API, Chris? Or can we shut this one down?($20/month)

linode_ddw_dev is running a native postgres on port 65432 and a  dockerized postgres remapped to 5433, but otherwise I cannot find any services running on it or think of anything dependent on it. ($40/month)

If the Spotlights are de-coupled from api.devinit.org, I should be good to go ahead and shut down linode1009616_ddw (dumping the database to Box for preservation, maybe), linode3673430_dh_staging, and linode_ddw_dev.
Alex Miller I Data Scientist

Sign up to our newsletter and topic updates

wakibi:
This exposes the Analyst UI DB over port 65432 so the Spotlights API can pick data off it.

### DIFF
--- a/Dockerfile_pg
+++ b/Dockerfile_pg
@@ -6,5 +6,15 @@ LABEL maintainer="akmiller01 <Alex Miller, alex.miller@devinit.org>"
 ENV POSTGRES_USER=analyst_ui_user
 ENV POSTGRES_PASSWORD=analyst_ui_pass
 
+# Copy custom configs
+COPY ./config/pg/postgresql.conf /etc/postgresql/postgresql.conf
+COPY ./config/pg/pg_hba.conf /etc/postgresql/pg_hba.conf
+
+RUN chmod 644 /etc/postgresql/postgresql.conf
+RUN chmod 644 /etc/postgresql/pg_hba.conf
+
+# Run DB with custom configs copied above
+CMD ["-c", "config_file=/etc/postgresql/postgresql.conf"]
+
 # Add initialization script to initdb
 ADD ./database_setup/dba.sql /docker-entrypoint-initdb.d/
--- a/Dockerfile_spotlights_api
+++ b/Dockerfile_spotlights_api
@@ -1,0 +1,23 @@
+# FROM node:9-alpine@sha256:1fbcd77d0eb2af765d24ae4758b5a94b0d55c6002a15a4431d55795a449ebd3d as clone
+FROM node:12-alpine as clone
+
+RUN apk update && apk add git
+RUN git clone https://github.com/devinit/ddw-external-api.git
+
+# End of intermediate build step
+
+# FROM node:9-alpine@sha256:1fbcd77d0eb2af765d24ae4758b5a94b0d55c6002a15a4431d55795a449ebd3d as api-release
+FROM node:12-alpine as api-release
+MAINTAINER vagrant <Chris Wakibi, chrisw@devinit.org>
+
+RUN mkdir /src
+COPY --from=clone /ddw-external-api /src/ddw-external-api
+COPY ./config/spotlights/db.conf /src/ddw-external-api/src/db.conf
+WORKDIR /src/ddw-external-api
+
+# The below fixes an issue with Github CI
+# RUN npm config set unsafe-perm true
+# update npm and install dependencies
+RUN npm i -g npm && npm install --silent
+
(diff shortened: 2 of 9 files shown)


k8hughes:
@akmiller01 could you take a look at this when you get online?

akmiller01:
@wakibi we'll need to change this in the Dockerfile_pg as well as in the web container settings.py. It will need to be a new username with a random password, set through environmental variables so they're not committed to the repo. Otherwise anyone who looks at our Github repo would be able to login to an exposed database: https://github.com/devinit/ddw-analyst-ui/blob/develop/Dockerfile_pg#L6-L7

wakibi:
Hi @akmiller01, I am managing that by creating another user that will be whitelisted in the postgresql.conf file, as well as only allowing connections to that port from the server running the API.

wakibi:
Deployment process via git;

- Pull latest git changes
- Switch to `feature/import_spotlights_data` branch
- Create directories `letsencrypt` and `spotlights` under the `static` folder in case they do not exist
- Copy the keys (`fullchain.pem` and `privkey.pem`) from the ssl folder on your local machine to the ssl folder on the server
- Copy the above keys to `spotlights_fullchain.pem` and `spotlights_privkey.pem` 
- Run npm install and npm run build
- Build the Docker containers
- Run `docker-compose up -d`
- Run `certbot certonly` as root, Choose option 3 (Place files in web-root)
- Input the sub-domain you are creating the certificate for i.e [staging-ddw.devinit.org](staging-ddw.devinit.org) or [staging-api.devinit.org](staging-api.devinit.org)
- Input the web-root i.e `/root/ddw-analyst-ui/static/letsencrypt` for ddw or `/root/ddw-analyst-ui/static/spotlights` for spotlights API
- Run `certbot renew --dry-run` to confirm the renewal process works

akmiller01 (approved):
Looks good!

Issue #421, opened 2021-02-03
New auto scheduled event run instances are created based on the previous run end time


edwinmp:
While the `date` should be based on the previous run's end date (in the event that a run goes on for too long), the `time` should be based on the previous run's start time. This will ensure start time consistency. Currently the start time gets later and later with each run.

k8hughes:
@davidebukali  I think I've seen a PR connected to this... is that right?
If so can you move this to code review and link the PR.

edwinmp:
This seem to be functional on staging
![image](https://user-images.githubusercontent.com/5672438/108235653-15658400-7157-11eb-8dbe-900450ed1962.png)

PR #430, opened 2021-02-11
Run scheduled events at the same time for interval types greater than hours

davidebukali:
#421

### DIFF
--- a/data/management/commands/run_schedules.py
+++ b/data/management/commands/run_schedules.py
@@ -114,12 +114,12 @@ class Command(BaseCommand):
 
             # This makes sure the next runtime is calculated from the time the last one was run i.e current time
             updated_runtime = self.update_next_runtime_to_previous_start_time(schedule, runInstance.start_at, make_aware(datetime.now()))
-            actual_runtime = self.is_updated_runtime_less_than_expected_runtime(updated_runtime, runInstance, schedule)
+            actual_runtime = self.get_most_recent_of_updated_runtime_or_expected_runtime(updated_runtime, runInstance, schedule)
             self.create_next_run_instance(schedule, actual_runtime)
         except:
             self.update_run_instance(runInstance, 'e', 'An unexpected error occured while executing the script ... please contact the administrator')
             updated_runtime = self.update_next_runtime_to_previous_start_time(schedule, runInstance.start_at, make_aware(datetime.now()))
-            actual_runtime = self.is_updated_runtime_less_than_expected_runtime(updated_runtime, runInstance, schedule)
+            actual_runtime = self.get_most_recent_of_updated_runtime_or_expected_runtime(updated_runtime, runInstance, schedule)
             self.create_next_run_instance(schedule, actual_runtime)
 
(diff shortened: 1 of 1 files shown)


edwinmp (changes requested):
I don't think u need to complicate it to handle illogical configs. It's as simple as each SE should have its start time calculated from the `start time` of the previous instance & the `configured interval`.
If user decides to be unrealistic & configures a SE to run every 30mins, yet each run lasts 2hrs, that's on them. We've already handled that by preventing concurrent runs of the same event.
What I expect here is to move the `create_next_run_instance` to when the current one starts.
Perhaps the only other validation you can do is to make sure that if the `Expected runtime` is configured, then be sure to set the new instance to start after the current one is expected to end.

data/management/commands/run_schedules.py
```diff
@@ -114,28 +114,42 @@ def run_and_update_schedule(self, schedule, runInstance):
...
-                minute=instance_start_date.minute,
-                second =instance_start_date.second
+        updated_runtime = next_runtime.replace(
+            hour=instance_start_date.hour,
+            minute=instance_start_date.minute,
+            second =instance_start_date.second
+        )
+        return updated_runtime
+
+    def is_updated_runtime_less_than_expected_runtime(self, updated_runtime, current_runInstance, schedule):
```
edwinmp: This naming implies that this returns a boolean...

edwinmp (changes requested):

data/management/commands/run_schedules.py
davidebukali: Updated function naming

data/management/commands/run_schedules.py
```diff
@@ -113,11 +113,42 @@ def run_and_update_schedule(self, schedule, runInstance):
...
+            updated_runtime = self.update_next_runtime_to_previous_start_time(schedule, runInstance.start_at, make_aware(datetime.now()))
+            actual_runtime = self.get_most_recent_of_updated_runtime_or_expected_runtime(updated_runtime, runInstance, schedule)
+            self.create_next_run_instance(schedule, actual_runtime)
+
+    def update_next_runtime_to_previous_start_time(self, schedule, instance_start_date, current_date_time):
+        next_runtime = self.calculate_next_runtime(
+            current_date_time,
+            schedule.interval,
+            schedule.interval_type
+        )
```
edwinmp: Correct me if I'm wrong, but I think we only need this for this function... just needs to replace `current_date_time` on line 127 with `instance_start_date`

edwinmp (commented):
Sorry to be giving you fragmented reviews ... will try to be thorough from the start next time.

edwinmp (dismissed):
If you're sure...

data/management/commands/run_schedules.py
davidebukali: You are right, using `instance_start_date` ensures we have the same start time for full cycle intervals (greater than 24hrs).

edwinmp (approved):

PR #444, opened 2021-02-22
Fix dev compose file to connect to Postgres

wakibi:


edwinmp (approved):
This works & maintains the separation until we upgrade the main settings to `Postgres 12`

PR #445, opened 2021-02-22
Reinitialise query builder form when values change

### LINKED ISSUE #447: Query builder issue name/description
There seems to be a bug on the UI where it is not noticing that a name and description have been added to a query, when it has, meaning the query can't be saved. I think its only applicable for when the 'is draft' box is not checked.

image

*Edit: now seems applicable only for when the 'is draft' box is checked.

davidebukali:
Query builder form shows validation errors even when field values are filled

Screenshot 2021-02-22 at 20 43 10


edwinmp (changes requested):
Tested this and it's fine, but it seems an ideal time for you to start writing some end-to-end tests. I've added the blank test (currently skipped), so just fill it up.

edwinmp (approved):

PR #446, opened 2021-02-22
Feature/test queries command

### LINKED ISSUE #428: Add QueryBuilder tests
- Testing of current queries to make sure they are not broken by changes to the querybuilder
- Testing of table column changes to make sure they do not break current queries

wakibi:
Should be able to run through all existing queries and determine whether changes to the QueryBuilder class will break them or not. Can also be able to determine whether changes to table columns will break any existing queries.

`python3 manage.py dry_run_current_queries -t=fts_recipientcountryid transaction_receiver_org_type disbursements_score_notes`

Takes 3 options;

- `-all` or `-a` option, specifies to run for all current queries
- `--table` or `-t` option, specifies which table the old columns we are renaming belong to
- All other options will be considered column names for the table specified with the `--table` option. This is a compulsory option (provide at least one column, even a dummy one) but will be ignored if no `--table` option is provided.

Output will be coloured for easy perusal.
On detection of an error, the user will be prompted to `Press the enter key to continue`

### DIFF
--- a/data/management/commands/dry_run_current_queries.py
+++ b/data/management/commands/dry_run_current_queries.py
@@ -9,18 +9,24 @@ from data.db_manager import analyse_query
 class RunCurrentQueries(BaseCommand):
 
     def add_arguments(self, parser):
-        parser.add_argument('active_mirror_name', type=str, help='Table name where the columns belong')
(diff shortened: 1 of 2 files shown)


akmiller01 (approved):
LGTM!

Issue #397, opened 2021-01-25
Add tooltips & help infrastructure that links to the Google documentation


edwinmp:
- [x] Add a nav bar item right after the `user` icon on the top right - should be explicitly named Help or have a help icon
- [x] Bind this link to the react router, changing it's destination with each page that's opened
- [x] Populate the dropdown with relevant links from the [User Guidance](https://docs.google.com/document/d/1j10m46dLY-zb5Mee4JWncTGeOsf2o8oChSN90ZUu8vk/edit#heading=h.b36pcn2w5m6j)

edwinmp:
@bill-anderson @Duncan-Knox I added a help menu to the top-right of each page. This menu changes for every page, with links to the Google documentation relating to the features on that page. Please test it out on [staging](https://staging-ddw.devinit.org/) ... same credentials as on live. 

![image](https://user-images.githubusercontent.com/5672438/109610285-9d776080-7b3d-11eb-969c-ce354bf8f5f7.png)

Duncan-Knox:
This looks fantastic @edwinmp, I've tested following the links through to the Google doc and all seems okay there. One question, I did notice the 'Query Builder' and 'Update Data Source' drop downs do not have a 'What you should see' link at the top whereas the other pages all did - flagging this for your consideration as not sure whether this choice was intentional?

PR #450, opened 2021-02-25
Feature | Help

edwinmp:
Issue #397 

Added a help menu to the top-right of the page. This menu changes based on the active page.
image

I think we'll need to move the configs for this into the data repo so the links can be updated without pushing code.


akmiller01 (dismissed):
Looks cool

PR #451, opened 2021-02-25
Alias creation bug fix

### LINKED ISSUE #370: Unable to access FTS data
I tried to perform several distinct queries on FTS data (query numbers 789, 790 and 791) but in every case they come up empty. Not sure what the issue is there.
The data search on the FTS website however also seems stuck and does not spit out data for me - might be related, not sure. 

Wondering, if the data pull from FTS is indeed broken (potentially due to issues on their side) - do we have frozen datasets from past data pulls we could use instead?

ThatcherK:
This branches handles the different stages involved in alias creation as well as a few minor bug fixes


frontend/src/pages/QueryBuilder/QueryBuilder.tsx
```diff
@@ -173,17 +173,29 @@ const QueryBuilder: FunctionComponent<QueryBuilderProps> = (props) => {
           data,
         })
         .then((response: AxiosResponse<Operation>) => {
-          if (response.status === 200 || response.status === 201) {
-            props.actions.operationSaved(true);
-            if (preview) {
-              props.history.push(`/queries/data/${response.data.id}/`);
-            } else {
-              props.history.push('/');
+          if (response.data.alias_creation_status !== 'p') {
```
edwinmp: This shouldn't be here ... it's not possible that this could happen since if the code runs without interruption, it can only have two status ... `d` & `e` (probably only `d`)... So in addition to the error check on line 191, this check should also be done in `setActiveOperationByID` (I think), to be run when the user first opens the query builder to edit an existing operation. i.e. `if (operation.alias_creation_status !== 'd')`. Alternatively, u can have a `useEffect` that watches for changes in `props.activeOperation`, then checks the alias creation status.

frontend/src/types/operations.ts
```diff
@@ -17,6 +17,7 @@ export interface Filters<T = Filter[] | ErroredFilter[]> {
   filters: T;
 }
 export interface Operation {
+  alias_creation_status: string;
```
edwinmp: Give it a more specific type i.e. `'d' | 'e' | 'p'`

edwinmp (changes requested):
🎩

edwinmp:
Your branch is behind develop by more than a few commits, and since you're going to need to add end-to-end testing for this feature, you'll need code that's on develop. So please rebase and push (you'll have to force push). If you're not confident with rebasing, create a new branch off `fix/alias`, do the rebase in there and push. Can review that first before u try it on this branch.

frontend/src/pages/QueryBuilder/QueryBuilder.tsx
ThatcherK: Oh yes,this sounds more practical

frontend/src/pages/QueryBuilder/QueryBuilder.tsx
```diff
@@ -102,6 +102,11 @@ const QueryBuilder: FunctionComponent<QueryBuilderProps> = (props) => {
   const setActiveOperationByID = (id: string) => {
     const operation = props.operations.find((ope) => ope.get('id') === parseInt(id, 10));
     if (operation) {
+      if (operation.get('alias_creation_status') === 'p') {
```
edwinmp: Check `!== 'd'` ... I'd rather flag this for error events as well.

edwinmp (changes requested):
As I said in a previous comment, you'll have to add Cypress tests to this. There's info in the READ me under end-to-end testing.

frontend/src/pages/QueryBuilder/QueryBuilder.tsx
ThatcherK: This is not very clear

edwinmp: Very well ... replace this with:
`if (operation.get('alias_creation_status') !== 'd') {`

ThatcherK: Okay

Dockerfile_pg_dev
```diff
@@ -1,5 +1,5 @@
 # start with a base image
-FROM postgres:12.5
+FROM postgres:11
```
edwinmp: Nope ... this may have fixed your issue but it'll break all ours. Upgrade your local postgres to the expected version instead. (soon to be version 13)

cypress.json
```diff
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "https://staging-ddw.devinit.org",
+  "baseUrl": "http://localhost/",
```
edwinmp: You need to be careful what you commit ... at the very least, review the files you stage before you commit them.

frontend/cypress/fixtures/users.json
```diff
@@ -1,10 +1,10 @@
...
   {
     "id": 1,
-    "name": "Admin McAwesome",
-    "username": "admin@test.com",
-    "email": "admin@test.com",
-    "password": "Spp9k6i7",
+    "name": "Admin",
+    "username": "Admin",
+    "email": "admin@mail.com",
+    "password": "atwookimomo@23",
```
edwinmp: This is going to break all tests on deployment ... instead of changing these to suit yours, why not create a user with the expected credentials. I mean, this didn't need to change at all really.

frontend/cypress/integration/queryBuilder.spec.js
```diff
@@ -45,6 +45,19 @@ describe('The Query Builder', () => {
...
 
+  it('that encountered an error or interruption during alias creation shows a warning', () => {
+    // Mock datasets route
+    cy.fixture('datasets').then((datasets) => {
+      cy.intercept('api/datasets/mine/', datasets);
+    });
+
+    cy.visit('/');
+    cy.get('.p-0 div a').first().click({ force: true });
+    cy.get('.p-1').contains(
```
edwinmp: These classes `p-0, p-1` are at high risk of changing, which will break your tests. I'd rather you add a testId to your elements.

frontend/cypress/integration/queryBuilder.spec.js
```diff
@@ -45,6 +45,19 @@ describe('The Query Builder', () => {
     cy.get('.invalid-feedback').should('not.be.visible');
   });
 
+  it('that encountered an error or interruption during alias creation shows a warning', () => {
+    // Mock datasets route
+    cy.fixture('datasets').then((datasets) => {
+      cy.intercept('api/datasets/mine/', datasets);
+    });
```
edwinmp, line 72: This was brilliant ...

edwinmp (changes requested):

frontend/cypress/fixtures/users.json
ThatcherK: I was getting an unauthorized error

edwinmp: I'm sure ... so create a user with the expected credentials, either via Django admin or with [createsuperuser](https://djangocentral.com/creating-super-user-in-django/) ... u only need the matching username & password.

edwinmp:
Also, there's conflicts with `develop`, so you'll need to rebase.

edwinmp (approved):
Great work on this ... learned a few things myself.

PR #459, opened 2021-03-03
Upgrade/postgres

wakibi:
Updates deployment process to use metadata2 volume as well as Postgres 13.2


Dockerfile_pg
```diff
@@ -1,5 +1,5 @@
 # start with a base image
-FROM postgres:11
+FROM postgres:13.2
```
edwinmp, line 2: Now that we're finally doing this, I don't think we need the `Dockerfile_pg_dev` file anymore ... think you can delete it and update `docker-compose.dev.yml`

edwinmp (dismissed):

edwinmp (approved):

PR #467, opened 2021-03-11
Feature/update iati publishers

### LINKED ISSUE #366: Update IATI Publishers Table during ETL
(based on our discussion yesterday)

During ETL

1. Check whether reporting-org_ref exists in the IATI Publishers Table
2. If not append it and reporting-org_narrative to table

wakibi:
Run;

`docker-compose exec web python manage.py update_csv_files -b=iati_reporting_org_narrative`
`docker-compose exec web ./data_updates/manual_data.sh`
`docker-compose exec web python manage.py update_iati_publishers_table`

Note the last command has been set up to run as part of the `iati.sh` script


data/management/commands/update_iati_publishers_table.py
```diff
@@ -0,0 +1,29 @@
...
+PUBLISHERS_TABLE_NAME = "iati_publishers"
+
+class Command(BaseCommand):
+    help = 'Downloads CSV files from git repo'
+
+    def handle(self, *args, **kwargs):
+        self.updateIatiPublishersTable()
+
+    def updateIatiPublishersTable(self):
+        engine = create_engine('postgresql://analyst_ui_user:analyst_ui_pass@db:5432/analyst_ui')
```
akmiller01: Using SQLalchemy within a management command feels a bit weird. Would it be possible to use the utilities from our db_manager? E.g. https://github.com/devinit/ddw-analyst-ui/blob/develop/data/db_manager.py#L6 ?

akmiller01 (changes requested):

data/management/commands/update_iati_publishers_table.py
wakibi: Actually all iati_transaction*.py scripts seem to take the route of SQLalchemy, so shall we have to update all of them?

akmiller01 (commented):
@wakibi difference is that all of the `data_update/Python/*.py` scripts are run outside of the Django environment, and therefore don't have immediate access to the database objects. Management command are run in the context of Django (all of the app gets loaded when the management command is invoked), so it can use database access the same way the rest of the website does.

wakibi:
Ah, why was I thinking they were under the commands subfolder? I shall simply move this to the data_updates sub-folder then.

data/management/commands/update_iati_publishers_table.py
wakibi: Moved it to the `data_updates` folder and updated the `iati_transactions.sh` script consequently.

akmiller01 (approved):
Looks good!

Issue #399, opened 2021-01-25
Show feedback on making breaking changes to a query builder step


edwinmp:
e.g. With 2 steps, one a selected & the next one a filter, if the select step is edited to remove the column being filtered in step 2, a warning should at least be shown about this.

k8hughes:
Has this been deployed?  if so we can move this to done

edwinmp:
It's on staging, going to be part of the v2.9.0 update.

k8hughes:
Ah ok.
@dean-breed it would be great if you could view this on staging and let us know if it is working as you'd expect.

dean-breed:
This appears to work well. No obvious problems

dean-breed:
@edwinmp @k8hughes this is all good to deploy.

PR #476, opened 2021-03-17
Feature/select_query_feedback

davidebukali:
Update users when they try to deselect a column used in another query step #399


edwinmp (changes requested):
Failed to build this, probably something on my local machine ... anyway, this `CheckboxGroup` is a presentational component and not meant to hold all that logic ... didn't look into it critically but I know it doesn't belong there.

frontend/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -6,6 +6,7 @@ interface ComponentProps {
   options: DropdownItemProps[];
   selectedOptions?: string[];
   onUpdateOptions?: (options: string) => void;
+  validateCheckboxOptions?: (option: string, checkboxState: boolean | undefined) => void;
```
edwinmp: I'm pretty sure you only care about the deselection of a checkbox, so lets rename this to `onDeselect` and remove `checkboxState` ... You'll just have to make sure you only call this if the checkbox is deselected.
Also, there's a much better way of writing `checkboxState: boolean | undefined`, without the `| undefined`

edwinmp: I also want the name to not be about validation, cos I may want to use this nice `CheckboxGroup` component outside of the QB, and do something else on deselect.

frontend/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -31,6 +33,8 @@ const CheckboxGroup: FunctionComponent<ComponentProps> = (props) => {
       ? checkboxes?.concat(data.value as string)
       : checkboxes?.filter((checkbox) => checkbox !== data.value);
     addCheckboxes(updatedCheckboxes);
+    props.validateCheckboxOptions &&
+      props.validateCheckboxOptions(data.value as string, data.checked);
```
edwinmp: This is where you check if the checkbox is deselected, then call the function. Also, we don't use this kind of syntax when calling functions. You put it in a proper if statement. More readable. It's the recommended syntax for assignment operations, even those involving function calls. Basically, I wouldn't complain if it was:
`const valid = props.validateCheckboxOptions &&
      props.validateCheckboxOptions(data.value, data.checked);`

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
+import { List } from 'immutable';
+import React, { FunctionComponent, useState } from 'react';
+import { Filters, OperationStepMap } from '../../types/operations';
+import { BasicModal } from '../BasicModal';
+
+interface ComponentProps {
+  steps?: List<OperationStepMap>;
+}
+
+const CheckboxGroupValidator: FunctionComponent<ComponentProps> = (props) => {
```
edwinmp: Given the previous comments, I'm sure you won't be surprised that I'm asking you to move this component to its own folder, and rename it to something not related to the `CheckboxGroup`. Name it something related to the `select` step since it'll be associated with it... e.g. `SelectColumnValidator`

edwinmp: You'll also need to have the active step as a prop, cos your current implementation doesn't take into account that a `select` step can come AFTER a filter. So it checks all filters, even those not affected by the changes in the current select. Let me know if you need more clarification on this.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const [showModal, setShowModal] = useState<boolean>(false);
+  const [modalMessage, setModalMessage] = useState<string>('');
```
edwinmp: I think the modalMessage state can be enough on it's own ... Unless u have a situation where it happens, if there's nothing to show, hide the modal.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  const validateCheckboxOptions = (
```
edwinmp: Rename this to `onDeselectColumn`.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+  ): void => {
+    const { steps } = props;
+    steps &&
+      steps.map((step) => {
+        const options = step.get('query_kwargs') as string;
+        toggleModalState(options, checkboxState, checkboxValue);
+      });
+  };
+
+  const toggleModalState = (
```
edwinmp: Rename this `validateColumnDeselect`

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+  steps?: List<OperationStepMap>;
+}
+
+const CheckboxGroupValidator: FunctionComponent<ComponentProps> = (props) => {
+  const [showModal, setShowModal] = useState<boolean>(false);
+  const [modalMessage, setModalMessage] = useState<string>('');
+
+  const validateCheckboxOptions = (
+    checkboxValue: string,
+    checkboxState: boolean | undefined,
```
edwinmp: You won't need this.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+  ): void => {
+    const { steps } = props;
+    steps &&
+      steps.map((step) => {
+        const options = step.get('query_kwargs') as string;
+        toggleModalState(options, checkboxState, checkboxValue);
+      });
```
edwinmp: Use an `if` statement.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
edwinmp: Also, as far I can tell, the only code related to the component that this function uses are state updates. That makes it a good candidate for extraction into a `utils` file (under the same component folder)`. Make it a promise.
i.e. What I expect to see here is:

`validateColumnDeselect(...).then(validationMessage => if (validationMessage) { setModalMessage(validationMessage); }`

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+  };
+
+  const toggleModalState = (
+    options: string,
+    checkboxState: boolean | undefined,
+    checkboxValue: string,
+  ): void => {
+    const parsedOptions = parseOptions(options);
```
edwinmp: I'm sure this `parseOptions` function will go into utils as well... Should leave us with a much cleaner component.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
edwinmp: This seems like the place to use the active step prop to filter these steps so that you only check those that come AFTER our active step.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+      }
+
+      return result;
+    }, []);
+  };
+
+  const parseOptions = (options: string) => {
+    const parsedOptions = JSON.parse(options);
+    if (parsedOptions.hasOwnProperty('filters')) {
+      const { filters }: Filters = options ? parsedOptions : { filters: [] };
```
edwinmp: I find this `options` check unnecessary ... since parsedOptions is created from `options` and you're checkin to make sure that `parsedOptions` has a filter property, then this particular check is redundant. Apply elsewhere.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+            break;
+          }
+        }
+      }
+
+      return result;
+    }, []);
+  };
+
+  const parseOptions = (options: string) => {
```
edwinmp: Give this a proper return type.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
edwinmp: I think this change is likely to get rid of all these strange default values you're adding... the whole thing looks like it was copied from somewhere.

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
```diff
@@ -0,0 +1,221 @@
...
+    <>
+      <BasicModal show={showModal} onHide={toggleShowModal}>
+        {modalMessage}
+      </BasicModal>
+      {React.Children.map(props.children, (child) => {
+        if (React.isValidElement(child)) {
+          return React.cloneElement(child, { validateCheckboxOptions });
+        }
+
+        return child;
```
edwinmp: Rewrite in a single statement.

edwinmp (changes requested):
This is a lot more comments than I've had to leave in a while, and some could have been avoided. It may help to read through all the comments before making the code changes... should save you some time.

frontend/src/components/CheckboxGroup/CheckboxGroup.tsx
davidebukali: Renamed to `onDeselect` and is only called when checkbox is deselected. Also removed `checkboxState`

frontend/src/components/CheckboxGroup/CheckboxGroup.tsx
davidebukali: Updated

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
davidebukali: Renamed and added active step prop

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
davidebukali: Updated

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
davidebukali: Moved to `utils` and updated to return a promise

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
davidebukali: Removed unnecessary parameter

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
davidebukali: Updated and added filter to exclude active step

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
davidebukali: Updated

frontend/src/components/CheckboxGroup/CheckboxGroupValidator.tsx
davidebukali: Updated

frontend/src/components/SelectColumnValidator/utils/index.ts
```diff
@@ -0,0 +1,116 @@
+interface ParsedOptions {
+  id: string;
+  values: any[];
+}
+
+export const validateSelectColumnDeselect = (options: string, checkboxValue: string) => {
```
edwinmp: Give this a return type ...

frontend/src/components/SelectColumnValidator/utils/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  return new Promise((resolve: (validationMessage: string) => void) => {
+    const parsedOptions = parseOptions(options);
+    parsedOptions?.values?.flat().reduce(function (result, query) {
+      if (query && query.field === checkboxValue && parsedOptions.id === 'filters') {
+        resolve(
+          `Notice that ${query.field
+            .split('_')
+            .join(' ')} is used in a filter step, deselecting it here may break this query.`,
```
edwinmp: Given the repetitive occurrences, extract this into a variable and keep switching out the dynamic bits i.e. field & step (filter, aggregate, e.t.c.)

frontend/src/components/SelectColumnValidator/utils/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    id = 'multi';
+  }
+  if (parsedOptions.hasOwnProperty('over')) {
+    id = 'window';
+  }
+
+  return {
+    id,
+    values,
+  };
```
edwinmp: Return within each if statement or use `else if` ... otherwise with the way you have it now, if more than one happen to be true, one will overwrite the other.

edwinmp (changes requested):

edwinmp:
Please start working on the end-to-end tests for this.

frontend/src/components/SelectColumnValidator/SelectColumnValidator.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+const SelectColumnValidator: FunctionComponent<ComponentProps> = (props) => {
+  const [modalMessage, setModalMessage] = useState<string>('');
+
+  const onDeselectColumn = (checkboxValue: string): void => {
+    const { steps } = props;
+    if (steps) {
+      steps
+        .filter((step) => step.get('step_id') !== props.step.get('step_id'))
```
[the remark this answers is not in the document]
davidebukali: Updated

frontend/src/components/SelectColumnValidator/SelectColumnValidator.tsx
```diff
@@ -0,0 +1,53 @@
...
+        .map((step) => {
+          const options = step.get('query_kwargs') as string;
+          validateColumnDeselect(options, checkboxValue);
+        });
+    }
+  };
+
+  const validateColumnDeselect = (options: string, checkboxValue: string): void => {
```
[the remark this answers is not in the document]
davidebukali: Removed redundant function

frontend/src/components/SelectColumnValidator/SelectColumnValidator.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const SelectColumnValidator: FunctionComponent<ComponentProps> = (props) => {
+  const [modalMessage, setModalMessage] = useState<string>('');
+
+  const onDeselectColumn = (checkboxValue: string): void => {
+    const { steps } = props;
+    if (steps) {
+      steps
+        .filter((step) => step.get('step_id') !== props.step.get('step_id'))
+        .map((step) => {
```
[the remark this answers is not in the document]
davidebukali: Updated

frontend/src/components/SelectColumnValidator/SelectColumnValidator.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <>
+      <BasicModal show={modalMessage} onHide={toggleShowModal}>
+        {modalMessage}
+      </BasicModal>
+      {React.Children.map(props.children, (child) => {
+        return React.isValidElement(child)
+          ? React.cloneElement(child, { onDeselect: onDeselectColumn })
+          : child;
+      })}
```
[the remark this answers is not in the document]
davidebukali: Removed return

frontend/src/components/SelectColumnValidator/utils/index.ts
davidebukali: Updated

frontend/src/components/SelectColumnValidator/utils/index.ts
davidebukali: Now using string literals

frontend/src/components/SelectColumnValidator/utils/index.ts
davidebukali: Updated

k8hughes:
@edwinmp to review

edwinmp (dismissed):

edwinmp:
@davidebukali end-to-end tests are failing on staging

edwinmp (approved):

PR #484, opened 2021-03-29
Feature/duplicate step

ThatcherK:


edwinmp (changes requested):
Good work. Some of my comments may be obsolete as I made a tiny change which I'm sure has broken the styling. Well, there's a comment on that somewhere as well.

frontend/src/components/OperationSteps/OperationSteps.tsx
```diff
@@ -50,15 +65,27 @@ const OperationSteps: FunctionComponent<OperationStepsProps> = (props) => {
...
+                    <StyledButton
+                      title="Duplicate"
+                      variant="link"
+                      size="sm"
+                      data-testid="step-duplicate"
+                      onClick={() => props.onDuplicateStep(step)}
+                    >
+                      <i className="material-icons">content_copy</i>
+                    </StyledButton>
+                  </div>
```
edwinmp: Your styling seems to only hide the buttons, but the entire row is always showing, which leaves extra padding when one isn't hovering on the button. Update the styling to show/hide the entire actions row, not just the buttons.


How it is:
image


How it should be:
image

k8hughes:
@edwinmp to review

edwinmp:
@ThatcherK I'm in the process of testing this on staging. Besides that, all that's left is to add documentation for it. The item on the dropdown should have a caption around the lines of `Duplicate a Step`
image

edwinmp (approved):

PR #488, opened 2021-04-06
Feature/query frozen table

wakibi:
This adds an entry to the sources table, as well as (corresponding entries) to the sourcescolumnmap table so queries can be build using the frozen data.

On deletion of the frozen table, all corresponding entries i.e queries, the source table as well as all dependencies are deleted from the DB.


k8hughes:
@wakibi and @akmiller01 can we work to get this deployed today?  What is still required for this to be release?

wakibi:
Only pending a review from @akmiller01

akmiller01 (approved):
Working well!

PR #498, opened 2021-04-14
Feature/order select columns

ThatcherK:


frontend/src/components/SelectColumn/SelectColumn.tsx
```diff
@@ -0,0 +1,29 @@
...
+    id: props.id,
+  });
+  const style = {
+    borderStyle: 'solid',
+    width: '400px',
+    margin: '2px',
+    cursor: 'grab',
+    transform: CSS.Translate.toString(transform),
+    transition: transition?.toString(),
+  };
```
edwinmp: Its called `styled-components`... even if it didn't exist, this is discouraged unless you're doing animations.

frontend/src/components/SelectColumn/SelectColumn.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { CSS } from '@dnd-kit/utilities';
+
+interface SelectColumnProps {
+  id: string;
+}
+
+const SelectColumn: FunctionComponent<SelectColumnProps> = (props) => {
+  const { attributes, listeners, setNodeRef, transform, transition } = useSortable({
+    id: props.id,
+  });
```
edwinmp, line 30: I think you can have this on a single line.

frontend/src/components/SelectColumnOrder/SelectColumnOrder.tsx
```diff
@@ -0,0 +1,71 @@
...
+    if (props.onUpdateColumns) {
+      props.onUpdateColumns(JSON.stringify({ columns: orderedColumns }));
+    }
+  }, [orderedColumns]);
+
+  const sensors = useSensors(
+    useSensor(PointerSensor),
+    useSensor(KeyboardSensor, {
+      coordinateGetter: sortableKeyboardCoordinates,
+    }),
```
edwinmp: One line

frontend/src/components/SelectColumnOrder/SelectColumnOrder.tsx
```diff
@@ -0,0 +1,71 @@
...
+  );
+  const handleDragEnd = (event: DragEndEvent) => {
+    const { active, over } = event;
```
edwinmp: Inline the destructuring on line 39

frontend/src/components/SelectColumnOrder/SelectColumnOrder.tsx
```diff
@@ -0,0 +1,71 @@
...
+    if (over) {
+      if (active.id !== over.id) {
```
edwinmp: I don't see why this can't be `if (over && active.id !== over.id)` ... shouldn't change the logic.

edwinmp (changes requested):
Really good code ... feature works nicely

If they're available, the columns captions should be the aliases not the original column names as shown below ... could be a problem on my laptop that aliases aren't generated, but just in case, wanted it out there. Naturally, the original column names should be used if aliases aren't available.
image

Now, lets style it ...

frontend/src/components/SelectColumn/SelectColumn.tsx
ThatcherK: Why is it discouraged?

frontend/src/components/SelectColumn/SelectColumn.tsx
ThatcherK: Prettier complains when I make it a one liner

edwinmp:
The code seems fine & I'll do a review of the styling once a rebase has been done (can't run it on my machine without it - an you have conflicts). Otherwise, the SelectColumn component will need unit tests, and as always, this entire feature will need end-to-end testing (goes without saying).

ThatcherK:
Here is a screenshot of the new style:
Screenshot from 2021-04-16 17-43-32

edwinmp (approved):

Issue #252, opened 2020-07-27
Add caching to improve the performance of data retrieval operations


edwinmp:
Benchmark and settle on the ideal service for our needs
https://aws.amazon.com/redis/
https://aws.amazon.com/elasticache/redis-vs-memcached/

https://www.digitalocean.com/products/managed-databases-redis/

edwinmp:
Front-end caching could do as well ... should guarantee that a query is only slow to run the first time & for when the cache has been cleared. Adding a `graphql` layer to the API could help with this. Otherwise, we could use [localforage](https://localforage.github.io/localForage/)

edwinmp:
Caching has only been added to the query data page, so this is still incomplete.

What's left:

- [ ] My Datasets
- [ ] Published Datasets

PR #499, opened 2021-04-15
Feature | Caching

davidebukali:
Cache dataset queries with localForage #252


frontend/src/pages/QueryData/QueryData.tsx
```diff
@@ -79,7 +81,17 @@ class QueryData extends React.Component<QueryDataProps> {
...
+        if (data) {
+          this.props.actions.setOperationData(fromJS(JSON.parse(data)), {
+            id,
+            limit: 10,
+            offset: 0,
+          });
+        } else {
+          this.props.actions.fetchOperationData({ id, limit: 10, offset: 0 });
+        }
+      });
```
edwinmp: This shouldn't be here ... it should be abstracted under the already existing data fetching/updating code ... so in this case, I only expect local forage code to be in the QueryData saga. i.e. before data is fetched via API, there should be a check for it in the local cache. You should catch the operation's `update` date/time to use it as a check for whether you should fetch new data or go via the cache.

edwinmp (changes requested):

edwinmp (approved):

Issue #296, opened 2020-08-18
Re-ordering steps in query builder


bill-anderson:
The order in which steps are run is important for the performance of the query. 

- When an edit is saved the step is moved to the end of the query.
- It isn't possible to insert a step

This means that you need to rebuild a query from scratch every time you want to edit it.

edwinmp:
This isn't probable ... each subsequent step is dependent on the previous steps... if you have 4 steps & edit the first one, it invalidates the other 3, so u'll need to change them all. There's no way around that. It's the same with inserting a step ... may as well get rid of all the other steps up until the point you wish to insert a new one.

At this point, I'd caution care when creating these queries - best to know what you want to do before diving in, cos as you can see, with this kind of high dependency, there's limited flexibility.

bill-anderson:
I can understand that under certain circumstances inserting a step could cause a problem. The use case I am referring to is building complex filters - for which this shouldn't be a problem, right?

Could we not facilitate reordering together with error messaging when users do something illogical? Even if the error only says "you've done something stupid".

I want to push on this as I think complex queries (primarily involving joins and filters) will become more frequent. Analysts will build these through trial and error - so they won't know exactly what they want to do before diving in.

edwinmp:
If it's enough to accompany it with errors & warnings, then we'll build it. I'll think on an intuitive design for it that fits well with what we already have. Now that the design is wider & more focused, it should be easier to add more functionality without crowding it.

k8hughes:
@edwinmp can we link this into the advanced query builder work?

edwinmp:
Sure, but it'd be nice to have in the current version as well.

k8hughes:
@dean-breed  can you test this?  @davidebukali will ping you when it is on staging...

dean-breed:
Sure - @davidebukali let me know when it is on staging!

davidebukali:
Alright @dean-breed we are making a release, will reach out when it's deployed.

davidebukali:
@dean-breed its now on [staging](https://staging-ddw.devinit.org)

dean-breed:
Thanks @davidebukali - this seems to work well to me!

davidebukali:
Great ... thanks @dean-breed

PR #504, opened 2021-04-26
Feature | Order Steps

edwinmp:
Issue #296 

1. Added 2 buttons to the [OperationStepView](https://github.com/devinit/ddw-analyst-ui/blob/feature/order-steps/frontend/src/components/OperationStepView/OperationStepView.tsx#L71) component. They're arrow buttons; move up & move down
2. Clicking the up button moves the step to previous position. The down button does the opposite.
3. The first step doesn't have an up button. The last step doesn't have a down button.

### DIFF
--- a/frontend/cypress/integration/queryBuilder.spec.js
+++ b/frontend/cypress/integration/queryBuilder.spec.js
@@ -156,7 +156,7 @@ describe('The Query Builder', () => {
     cy.get('[data-testid="qb-order-step-button"]').should('be.visible');
   });
 
-  it('navigates to reorder step view', () => {
+  it('navigates to reorder steps view and back to normal steps view', () => {
     // Visit query builder, type name and choose datasource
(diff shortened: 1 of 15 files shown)


edwinmp:
@davidebukali you opted for this path so u are really going to have to build it right ... like properly right, not just kinda right ...

1. I said this before - sort button should only show up when it has value i.e when there's more than 1 step.
2. The entire row will need to move when dragging, not just some tiny outline with a number - otherwise, what was the point of ditching your initial implementation when this looks worse?
3. The info alert is not required here, especially since it's not helpful as it instructs users to drag anywhere on the entire step instead of using the handle.

edwinmp (commented):
@davidebukali I've updated the styling. Code looks good. Only 2 things left.

1. End-to-end tests
2. Documentation entry for this feature. There should be an appearance in the help menu that links to the doc

image

Issue #422, opened 2021-02-09
Update end-to-end testing to include core features


edwinmp:
- [x] Login
- [x] Query Builder (each function)
- [x] View query data
- [ ] Update data source

edwinmp:
As with the documentation, updates to this will need to be made as we go. They'll especially be required on PRs that affect core features.

edwinmp:
Adding end-to-end tests is now part of our expected deliverables for all work.

edwinmp:
@ThatcherK doing this for viewing query data is a high priority. Also the recent feature @wakibi added requires end-to-end testing (issue #470.

edwinmp:
@wakibi will you be adding the end-to-end tests for #470 yourself or should Thatcher do those as well?

edwinmp:
@ThatcherK I've added APIs to the checkbox list in the issue's main comment as one of the deliverables ...

wakibi:
> @wakibi will you be adding the end-to-end tests for #470 yourself or should Thatcher do those as well?

@ThatcherK can enjoy that one

ThatcherK:
Noted

edwinmp:
> @ThatcherK I've added APIs to the checkbox list in the issue's main comment as one of the deliverables ...

I've removed this.

edwinmp:
@ThatcherK 
The freeze feature in question is found under:
`Data Sources` -> `Versions` -> `Freeze Current Version`
and
`My Datasets` or `Published Datasets` -> `Versions` -> `Freeze Current Version`

They'll naturally need to be tested separately as they have slightly different use cases.

The data sources freeze feature in issue #470 allows the created frozen dataset to appear as a data source, so you'll need to verify that it exists under the `Data Sources` page and that you can query it as you would a normal data source... I'd prefer to avoid interception here as much as possible as we'll be testing everything from creation to deletion.

The datasets freeze feature only supports creation, download & deletion (I think).

Manually test these to see what to expect.

edwinmp:
@ThatcherK for the `Export to CSV` button:

1. Test that the button is visible and is a submit button
3. Test that the form has correct link in its action property and the correct method

![image](https://user-images.githubusercontent.com/5672438/116529253-6aefa700-a8e5-11eb-9326-4063a457a896.png)

This should be enough for now.

Reference:
[...]

PR #508, opened 2021-04-27
Update | End-to-End Testing

ThatcherK:
Issue #422


frontend/cypress/integration/datasets.spec.js
```diff
@@ -113,4 +113,75 @@ describe('The Datasets Pages', () => {
...
+      .should('be.visible')
+      .type('Test dataset freeze');
+    cy.get('[data-testid="dataset-frozen-data-save-button"]').should('be.visible').click();
+    cy.get('[data-testid="dataset-frozen-data-refresh-button"]').first().click({ force: true });
+    cy.get('[data-testid="dataset-frozen-data-status"]').contains('Completed');
+
+    // Download frozen dataset
+    const localStoragePrefix = 'ddw-analyst-ui/ddw_store/';
+    const token = window.localStorage.getItem(`${localStoragePrefix}API_KEY`);
+    if (token) {
```
edwinmp: Use command

frontend/src/components/DatasetHistoryListItem/DatasetHistoryListItem.tsx
```diff
@@ -33,7 +33,7 @@ export const DatasetHistoryListItem: FunctionComponent<ComponentProps> = (props)
           console.log(err.message);
         });
     }
-  }, [fetching]);
+  }, [fetching, item]);
   useEffect(() => setItem(props.item), [props.item]);
```
edwinmp: Instead of watching `item` on line 36 (wouldn't be useful if `fetching` is false), why not add `setFetching(true)` to the effect on line 37? Let me know if this change gives you the desired result in your tests.

frontend/src/components/SourceHistoryListItem/SourceHistoryListItem.tsx
```diff
@@ -32,7 +33,7 @@ export const SourceHistoryListItem: FunctionComponent<ComponentProps> = (props)
         })
         .catch((error) => console.log(error.message));
     }
-  }, [fetching]);
+  }, [fetching, item]);
```
edwinmp: With this one too.

edwinmp (changes requested):
Since you're busy making changes to the code, why not also fix that `Confirm Delete` bug you mentioned?

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -56,4 +56,86 @@ describe('The Data Sources Page', () => {
...
+          cy.get('[data-testid="frozen-data-form-message"]').should('be.visible').type('test');
+          cy.get('[data-testid="frozen-data-form-save-button"]').should('be.visible').click();
+          cy.get('[data-testid="frozen-dataset-refresh-button"]').first().click({ force: true });
+        });
+      cy.get('[data-testid="frozen-data-status"]').contains('Completed');
+
+      // Download frozen data source
+      const localStoragePrefix = 'ddw-analyst-ui/ddw_store/';
+      const token = window.localStorage.getItem(`${localStoragePrefix}API_KEY`);
+      if (token) {
```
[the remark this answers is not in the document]
ThatcherK: Noted

frontend/src/components/DatasetHistoryListItem/DatasetHistoryListItem.tsx
ThatcherK: The change gives me the desired results !

frontend/src/components/SourceHistoryListItem/SourceHistoryListItem.tsx
ThatcherK: It seems not to work for data sources though

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -56,4 +56,86 @@ describe('The Data Sources Page', () => {
         });
       });
   });
+
+  describe('data source update', () => {
+    it('freezes and deletes a data source', () => {
```
edwinmp: I'd rather you split them into separate tests ... that way, it's very clear which one is broken.

You'll also need to add a 3rd test for the info button in the even that the freezing fails ... refer to screenshots below. Feel free to intercept this one.

image

On clicking info, shows a modal:
image

edwinmp: The screenshots are for a datasets test, but the behaviour is the same for data sources. You'll naturally need to implement this for both.

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -56,4 +56,86 @@ describe('The Data Sources Page', () => {
...
+        .then(($element) => {
+          cy.wrap($element).contains('Versions').click();
+          cy.get('[data-testid="source-version-freeze-button"]').click();
+          cy.get('[data-testid="frozen-data-form-message"]').should('be.visible').type('test');
+          cy.get('[data-testid="frozen-data-form-save-button"]').should('be.visible').click();
+          cy.get('[data-testid="frozen-dataset-refresh-button"]').first().click({ force: true });
+        });
+      cy.get('[data-testid="frozen-data-status"]').contains('Completed');
+
+      // Download frozen data source
```
edwinmp: Extract into own test

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -56,4 +56,86 @@ describe('The Data Sources Page', () => {
...
+                `/api/tables/download/${currentFrozenDataSource.frozen_db_table}/archives/`,
+              );
+          });
+        }
+      });
+
+      // Check if new frozen data source can  be accesses like others
+      cy.visit('/sources');
+      cy.get('[data-testid="sources-table-search"]').type('FTS ISO codes test{enter}');
+      cy.get('[data-testid="sources-table-row"]').first().contains('FTS ISO codes test');
```
edwinmp: Extract into own test

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -56,4 +56,86 @@ describe('The Data Sources Page', () => {
...
+
+      // Query frozen data source
```
edwinmp: Extract into own test

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -56,4 +56,86 @@ describe('The Data Sources Page', () => {
...
+        .then(($element) => {
+          cy.wrap($element).contains('Versions').click();
+          cy.get('[data-testid="frozen-data-description"]').first().contains('test');
+          cy.get('[data-testid="frozen-data-delete-button"]').first().dblclick({ force: true });
+        });
+
+      //  Confirm that frozen data source version was deleted
+      cy.visit('/sources');
+      cy.get('[data-testid="sources-table-search"]').type('FTS ISO codes test{enter}');
+      cy.get('.row').contains('No Data');
```
edwinmp, line 158: Yeah, this is should be in, but before this, we'll need to know that the frozen record isn't showing up in the versions table ... I understand the previous challenge arose cause you were testing by name. So instead of that, verify that the number of frozen records is `X` in the beginning, and is `X - 1` after deleting.

frontend/cypress/integration/datasets.spec.js
```diff
@@ -111,6 +112,78 @@ describe('The Datasets Pages', () => {
...
+              cy.wrap($form).should('have.attr', 'action', '/api/export/4/');
+            });
+            cy.get('[data-testid="dataset-export-button"]')
+              .should('be.visible')
+              .should('have.attr', 'type', 'submit');
+          });
+      });
+  });
+
+  it('freezes and deletes a dataset', () => {
```
edwinmp: Apply all feedback on the data source tests to this one as well.

frontend/cypress/integration/datasets.spec.js
```diff
@@ -111,6 +112,78 @@ describe('The Datasets Pages', () => {
...
+              'href',
+              `/api/tables/download/${currentFrozenDataset.saved_query_db_table}/dataset/`,
+            );
+        });
+      }
+    });
+
+    // Delete frozen dataset
+    cy.get('[data-testid="frozen-dataset-delete-button"]').first().dblclick({ force: true });
+    cy.get('.p-0').contains('Test dataset freeze').should('not.exist');
```
edwinmp: This will fail when there are more than one records with duplicate names... count rows instead.

frontend/src/components/SourceHistoryListItem/SourceHistoryListItem.tsx
edwinmp: I'm hoping the new approach would eliminate the need for this change. If not, don't mind it.

edwinmp (changes requested):
Needed to rebase to test this properly, and I've pushed the rebased version ... Made some tiny changes too. I hope u won't have any issues on pull.

edwinmp (changes requested):
It appears that some of my previous feedback was missed. What helps is to go through each feedback item and comment that you've done it, that way you'll not likely miss anything.

edwinmp (changes requested):
Alright, seems something is getting lost in translation hence the back & forth (though we're getting there).

I need the tests to be organised logically so that we don't have stray test data left behind. i.e.

1. Freeze a dataset/source
2. Anything that needs to be tested on that frozen dataset/source that is NOT deleting
3. Delete that dataset/source

It's important that the delete happens last, otherwise we're left with random test data after each run & prolong other tests by having to create new data. I'd rather that those mid-level tests automatically PASS when the create dataset/source test fails. i.e. if you can't find rows to test with when you expect them, call the `done` function to automatically pass the test.
As for the delete, please delete everything in the table (if there's more than one).

edwinmp (approved):

PR #509, opened 2021-04-28
Feature/cache-datasets

davidebukali:
Cache user and published datasets


edwinmp (changes requested):
There's one MAJOR bug with this whole cache thing (including the query data cache) ... it doesn't take into account recent changes e.g. if I edit my dataset to return an entirely new set of data, the cache doesn't get updated so the data returned will be old.
It's the same in this case ... if I edit my dataset or create a new one, those changes won't show up cos the cache doesn't get update.

edwinmp:
Since the dataset data bug made it into the release branch, I'll fix that ... besides, going to be making extensive code changes to that page so we'll conflict if you work on it too.

edwinmp (approved):

Issue #473, opened 2021-03-17
Individual links for datasets & data source pagination & filters


edwinmp:
When paginating on datasets & data source lists, we'd like each pagination to update the URL query for that page i.e. `page=2`. That way, on navigating away from the page by clicking a button, going back should return the user to the correct place in the pagination.

![image](https://user-images.githubusercontent.com/5672438/111460169-b6326980-872c-11eb-8d73-7104421949c4.png)

Same for filters, both search and select.

PR #519, opened 2021-05-17
Feature | Individual Pagination Links

ThatcherK:
Page numbers appear in the url at the moment but on refresh, the data shown is reset to the first page ,although the url and pagination button show the correct page.

Issues #473


edwinmp:
This is incomplete, and implemented wrong (where it exists).

As far as I can see, you only have this for pagination yet I recall a question on what `filters` meant. Please implement this for all filters as well.

As to why I say this is implemented wrong, it's because the current implementation won't work when you account for filters. It can't be with new routes like you have it. These filters & pagination should be implemented via query parameters. So I expect urls to look like:

Pagination: `https://staging-ddw.devinit.org/sources/?page=1`
Filters: `https://staging-ddw.devinit.org/sources/?q="FTS"&source=12`

It can even handle all: `https://staging-ddw.devinit.org/sources/?page=1&q="FTS"&source=12` i.e. returns the first page of the results where the source ID is 12 and the name/description has `FTS` in it

edwinmp:
Also, the initial page is page 1 ... so any subsequent pages must be additive i.e 2, 3, 4 ... your pagination is `Page Number - 1`, so if I click page 3, the url is page 2. This is confusing.

edwinmp:
@ThatcherK this is also happening when I try to search data sources.
image

edwinmp:
@ThatcherK 
Functional Review Feedback

- [ ] Clicking the back for forward buttons on the browser only changes the URL but not the view i.e. if on page two and click back, URL changes to page 1 but the table is still showing page 2.

- [ ] Not sure why or if it's a bug you introduced, but filtering by data source seems to be broken ... I'll have to redeploy develop to determine the source

- [ ] Filters stay when they're not required anymore i.e. if I search or filter by data source, then press the back button, the table is reset but the search text or data source filter remains.

- [ ] When I'm on page 2, then I search/filter, and then press the back button (expecting to return to a page 2 filtered table), the URL is accurate but the data is for page 1

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
```diff
@@ -57,17 +58,25 @@ const getSourceDatasetsLink = (
   }${sourceID}?limit=${limit}&offset=${offset}&search=${search}`;
 
 const OperationsTableCard: FunctionComponent<OperationsTableCardProps> = (props) => {
+  const { search } = useLocation();
+  const queryParams = queryString.parse(location.search);
```
edwinmp: instead of `location.search`, use the `search` provided in the previous line... in fact, `useLocation()` eliminates the need for `window.location` so clear it out anywhere else.

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
```diff
@@ -57,17 +58,25 @@ const getSourceDatasetsLink = (
...
+  const [searchQuery, setSearchQuery] = useState((queryParams.q as string) || '');
   const [info, setInfo] = useState('');
   const [dropDownValues, setDropDownValues] = useState<DropdownItemProps[]>([]);
   const onModalHide = () => setInfo('');
   const { sources } = useContext(SourcesContext);
-
+  const [pageNumber, setPageNumber] = useState(Number(queryParams.page as string) || 1);
+  useEffect(() => {
+    if (Object.entries(queryParams).length === 0) {
+      props.history.push(`${window.location.pathname}?page=1`);
```
edwinmp: Use pathname from `useLocation()` hook

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
```diff
@@ -28,6 +28,7 @@ import { DatasetActionLink } from '../DatasetActionLink';
 import { OperationsTableRow } from '../OperationsTableRow';
 import OperationsTableRowActions from '../OperationsTableRowActions';
 import { PaginationRow } from '../PaginationRow';
+import queryString from 'query-string';
```
edwinmp: Organise imports ... VS code has this feature

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
```diff
@@ -57,17 +58,25 @@ const getSourceDatasetsLink = (
...
+      props.history.push(`${window.location.pathname}?page=1`);
+      setPageNumber(1);
+    }
+    setPageNumber(Number(queryParams.page as string));
+  }, [queryParams]);
   useEffect(() => {
     fetchQueries(showMyQueries);
-  }, []);
-
+  }, [pageNumber, search]);
```
edwinmp: Why watch these when you're also watching the query params? Seems redundant ... don't mind it though ... I feel a lot of this is inefficient but you've been on it too long & it needs to be closed.

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
```diff
@@ -110,15 +119,11 @@ const OperationsTableCard: FunctionComponent<OperationsTableCardProps> = (props)
...
-        mine: showMyQueries,
-        link: props.sourceID
-          ? getSourceDatasetsLink(props.sourceID, showMyQueries, props.limit, 0, value)
-          : undefined,
-      });
+      setPageNumber(1);
+      const values = queryString.parse(location.search);
+      values.q = value || null;
+      values.page = '1';
+      props.history.push(`${window.location.pathname}?${queryString.stringify(values)}`);
```
edwinmp: I see this pattern a few times ... page number, values, history push ... sure you can extract it into a reusable function

frontend/src/components/SourcesTableCard/SourcesTableCard.tsx
```diff
@@ -8,6 +8,8 @@ import { SourceMap } from '../../types/sources';
 import { PaginationRow } from '../PaginationRow';
 import { SourcesTable } from '../SourcesTable/SourcesTable';
 import { fetchSources } from '../../actions/sources';
+import { useHistory, useLocation } from 'react-router-dom';
+import queryString from 'query-string';
```
edwinmp: Organise imports

frontend/src/layouts/MainLayout.tsx
```diff
@@ -236,6 +236,7 @@ class MainLayout extends React.Component<MainLayoutProps, MainLayoutState> {
               <Route path="/" exact component={AsyncHome} />
               <Route path="/datasets" exact component={AsyncPublishedDatasets} />
               <Route path="/sources" exact component={AsyncDataSources} />
+              <Route path="/sources/page/:id" exact component={AsyncDataSources} />
```
edwinmp: Don't think this is required

edwinmp (changes requested):

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
ThatcherK: The queryparams are being watched in a separate useEffect

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -22,6 +22,40 @@ describe('The Data Sources Page', () => {
     cy.get('.sources-table').find('tbody').find('tr').should('have.length.lessThan', 11);
   });
 
+  it('shows correct url on pagination of data sources', () => {
```
edwinmp: This could be more general to cover all expected behavior on pagination ... i.e.
`it('paginates when datasets exceed 10')`

edwinmp: I'd expect to include a check on the URL and the results to make sure it actually paginated. Ofc you'd need to first verify that this ONLY happens when results exceed i.e. you'll need to render a table with less than 10 and verify that there's no pagination, then another with more than 10 to verify the reverse.

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -22,6 +22,40 @@ describe('The Data Sources Page', () => {
...
+            cy.url().should('eq', `${Cypress.config('baseUrl')}/`);
+          } else if (index < $linkCount && index !== 0) {
+            cy.get('.page-link').eq(index).click();
+            cy.url().should('eq', `${Cypress.config('baseUrl')}/?page=${index}`);
+          }
+        });
+      });
+  });
+
+  it('shows correct url and pagination when data sources are filtered by text', () => {
```
edwinmp: Simply `it('filters by text')`, then check all behaviour related to that.

frontend/cypress/integration/dataSources.spec.js
```diff
@@ -22,6 +22,40 @@ describe('The Data Sources Page', () => {
...
+    cy.url().should('eq', `${Cypress.config('baseUrl')}/sources/`);
+    cy.get('[data-testid="sources-table-search"]').type('code{enter}');
+    cy.url().should('eq', `${Cypress.config('baseUrl')}/sources/?page=1&search=code`);
+    cy.get('[data-testid="sources-table-search"]').should('have.value', 'code');
+    cy.go('back');
+    cy.url().should('eq', `${Cypress.config('baseUrl')}/sources/`);
+    cy.get('[data-testid="sources-table-search"]').should('have.value', '');
+  });
+
   describe('sources table', () => {
```
edwinmp: I expect all the tests above to be under this since they're all tied to the sources table.

frontend/cypress/integration/datasets.spec.js
```diff
@@ -93,6 +93,51 @@ describe('The Datasets Pages', () => {
     });
   });
 
+  it('shows correct url on pagination of datasets', () => {
```
edwinmp: Same as with data sources

frontend/cypress/integration/datasets.spec.js
```diff
@@ -93,6 +93,51 @@ describe('The Datasets Pages', () => {
...
+            cy.url().should('eq', `${Cypress.config('baseUrl')}/`);
+          } else if (index < $linkCount && index !== 0) {
+            cy.get('.page-link').eq(index).click();
+            cy.url().should('eq', `${Cypress.config('baseUrl')}/?page=${index}`);
+          }
+        });
+      });
+  });
+
+  it('shows correct url and pagination when datasets are filtered by text', () => {
```
edwinmp: Same as with data sources

edwinmp (changes requested):

edwinmp (approved):

Issue #531, opened 2021-05-26
Advanced Query Builder Filter


edwinmp:
We currently have the UI below:
![image](https://user-images.githubusercontent.com/5672438/119632342-630a2080-be19-11eb-921e-9cc43c2c3c96.png)

What's Missing:

- [x] Reset button - returns the filter editor to the default (depending on the selected AND/OR)
- [x] On adding a column filter, the value should be cleared
- [x] On replace/insert, validate the JSON before applying to main editor
- [ ] Clear all the tasks in the comments below

edwinmp:
> * On adding a column filter, the value should be cleared

This one may not be clear ... In the screenshot below, when a filter is added by clicking the add button, the value field should be cleared & ready to take in new input.

NB: uploading screenshots has failed, so ask for a call so I can share my screen

edwinmp:
We're going to need a new design for this ... the current one isn't as user friendly as I expected ... I'm open to suggestions. @davidebukali I remember you mentioned some kind of nested widget.

**TODO**

- [x] Design more intuitive UI for the filter helper
- [x] Implement filter helper UI

edwinmp:
**BUG**

- [x] Filter validation doesn't not account for complex filter configs like the one below & instead returns a bunch of errors

```
{
  "$or": [
    {
      "$and": [
        {
          "column": "Donor",
          "comp": "$eq",
          "value": "Central Emergency Response Fund"
        },
        {
          "column": "Recipient.Organization",
          "comp": "$eq",
          "value": "World Food Programme"
        }
      ]
    },
    {
      "$and": [
        {
          "column": "Donor",
          "comp": "$eq",
          "value": "Japan, Government of"
        },
        {
          "column": "Recipient.Organization",
          "comp": "$eq",
          "value": "World Food Programme"
        }
      ]
    }
  ]
}
```

edwinmp:
**BUG**

- [x] When the data source changes, any existing filter properties must be cleared

edwinmp:
We're going to customise and use [this](https://querybuilder.js.org/demo.html) query builder for the filter.

edwinmp:
We still have a PR open for this ... the updated design should help with ticket #580

edwinmp:
**Behaviour Updates**

- [ ] Sort columns alphabetically

**What's Missing**

- [ ] Not Equal
- [ ] Less Than or Equal
- [ ] Great Than or Equal
[...]

PR #536, opened 2021-06-02
Feature | Advanced QB Filter Updates

davidebukali:
Updates for advanced query builder filters #531


edwinmp (changes requested):
Good work on this.

image

There's an `x` that doesn't do anything ... that's a close button, so I hope you weren't using it as an error indicator ... convention makes us want to click it ... it's fine to simply remove it.

Rename the error message to `The operator $e is invalid` ... since we can have multiple configs for the same column, doesn't help much to mention it. In a later iteration of this, we shall be able to show these errors directly in the editor.

image

Finally, the requirement was for ONLY the value to be reset on adding a new filter ... by value, I meant he `value` field. This is reseting the whole thing.

edwinmp:
@davidebukali 

We need to make sure that the added filter config actually has something to filter. Currently, you can add a filter that's an empty array ... shouldn't be allowed

I've realised we are going to need on more button ... the `Remove/Clear` button ... sounds similar to reset but this one removes filter configs from the main JSON. i.e. the entire config is removed, even the `filter` property.

davidebukali:
@edwinmp added a clear button below the main JSON to remove all filters and included validation to prevent adding a filter without all values set.

edwinmp:
@davidebukali what's the hold up with this one?

Issue #520, opened 2021-05-17
Add Advanced Query Builder Select Actions


edwinmp:
While the JSON can be edited manually, `ACTIONS` are helpers that aid in this editing, reducing on errors. 

The Select `Action` adds the following properties to the QB JSON:

1. `selectAll` - boolean
2. `columns` - object array - if `selectAll` is true, this will be used to determine the column order

**Actions**:

- [x] Select Columns - used to select one or more columns. This action replaces all existing contents of the `columns` property with the selected columns.
- [x] Select ALL - a checkbox to toggle the `selectAll` property
- [x] Clear/Reset - Removes select properties
- [x] Aggregate - helps with aggregating selected columns. Will limit columns to number columns

Lastly, validate all select configs i.e.

- [ ] All columns must have valid IDs and names
- [ ] All columns must have an alias

![image](https://user-images.githubusercontent.com/5672438/118520073-5feca180-b742-11eb-8844-1720d7e790d5.png)

edwinmp:
@ThatcherK  For aggregation, the design should be similar to the filter UI shown below

![image](https://user-images.githubusercontent.com/5672438/122059528-66c80c00-cdf5-11eb-8090-0ab6e1e6061d.png)

Except it'll be just two drop-downs, one for selecting a column & the other for selecting the aggregation. The plus will stay, to add the aggregation.

So the behaviour is as follows:

- The aggregate button should only show if there are selected columns
- On clicking the aggregate button, the aggregate widget above is displayed
- The aggregate widget's select dropdown must only show selected columns i.e. columns that were chosen via the select action
- When the plus button is clicked (after selecting the column & aggregate), append the selected aggregate to the respective column in the JSON config.

Questions?

edwinmp:
Depends on issue #554 before implementing validation ... completed features have been merged into #478

edwinmp:
**New Features**

1. On reviewing error messages, it was brought to my attention that when a SQL statements has an aggregation, then the returned columns must either be aggregated or part of the `groupby` clause.

Therefore, if some of the selected columns are NOT aggregated, the default should be to add them to the `groupby` property.

![image](https://user-images.githubusercontent.com/5672438/123621214-a8a07b80-d813-11eb-8b10-b8ac5ff9b61e.png)


2. Only show the aggregate button when at least one of the selected columns supports aggregation i.e. is numerical

edwinmp:
**Bug**

- [ ] Ordering columns removes aggregation

PR #548, opened 2021-06-15
Feature | Advanced QB Select Actions

ThatcherK:
Issue #520


edwinmp (changes requested):
**Functional Feedback**

- Seeing as `select all` is not a button, it shouldn't be placed within a button group. It should appear on its own line, above the buttons.

image

- As is stated in the issue description, the select all is a checkbox ... this is a radio button so I can't uncheck it if I wanted to.

- As per the issue description, the select all should toggle the `selectAll` property. This means that the JSON should have a property `selectAll` that's either true or false based on the checkbox ... it is NOT supposed to auto select all columns. Please consult the screenshot below, from the issue description & let me know if there's anything u need clarity on.
image

**Additional Behaviour**


- [ ] Select All should true by default
- [ ] Clear/Reset should return the selectAll property to its default true **[Edit]:** this should also reset the `columns` property to an empty array
- [ ] When select all is true, the `Select Column(s)` button should be named `Select Columns for Ordering` ... no so sure about this, but let's give it a go & see. I do know that `Order Columns` wouldn't be accurate.

edwinmp:
https://github.com/devinit/ddw-analyst-ui/pull/548#pullrequestreview-685358953
@ThatcherK  All the above feedback remains as it is, so please go through it and make sure you've done it & then request a review **before** implementing this next one.

1. Add a new button `Order Columns`
2. Order Columns is disabled by default
3. Order Columns becomes enabled when more than 1 columns have been selected for ordering
4. Only selected columns are shown & ordered in the Order Columns drag & drop view
5. The new order must be reflected in the `columns` JSON property i.e ordering columns rearranges them in the `columns` property (currently its doing nothing)

edwinmp (approved):

PR #549, opened 2021-06-15
Feature | Query Estimate

### LINKED ISSUE #260: Progress when running a query.
1. Is it possible to show a progress bar when running a query?
2. Is it possible to show the total record count when viewing data prior to export?

wakibi:
Issue #260


edwinmp:
@wakibi please verify that this behaves as you expect on staging ... compare with the current behaviour and see if it's an improvement ... I had trouble seeing the expected root behaviour i.e. the query estimate. Let me know how to replicate it. My worry is that we've now optimised the behaviour so well that there's barely a delay, especially since we paginate the results. Anyway, let me know if you find a particularly slow query where the true value of this feature can be demonstrated.

edwinmp (changes requested):
This isn't doing what its supposed to, even for new queries. It's also adding overhead that's slowed things down a bit. We can fix this & the UX in subsequent reviews, but the main thing is that its not behaving as expected.

Issue #528, opened 2021-05-24
Process advanced QB configs to generate queries


edwinmp:
Complete the `advanced.py` module for handling the actual `advanced_config`. There's a sample config at the bottom of that file. There's info on the updates made to in this issue below. It's pretty complete for filters, select & joins (as far as I know - and I'm not a DB guru), so feel free to work out all the edge cases. This config is bound to change as we build the Advanced QB, with mostly additions.

We'll need to add to it to handle other query operations.

Current Config:

```
{
        'source': 1,
        'selectAll': True
        'columns': ['donor_name', 'agency_name'],
        'filter': {
            '$and': [
                { 'column': 'donor_name', 'comp': '$eq', 'value': 'United States' },
                { 'column': 'agency_name', 'comp': '$eq', 'value': 'Department of Agriculture' },
                {
                    '$or': [
                        { 'column': 'agency_name', 'comp': '$eq', 'value': 'Department of Agriculture' },
                        { 'column': 'agency_name', 'comp': '$eq', 'value': 'Miscellaneous' },
                    ],
                },
            ],
        },
        'join': {
            'source': 5,
            'type': 'inner',
            'mapping': [['donor_name', 'donor_name'], ['year', 'year']]
        }
    }
```

On the FE, the `columns` property has changed from the string array to an object array with the objects in the following format:

```
{
        id: 12,
        name: 'donor_name',
        alias: 'Donor Name',
        aggregate: 'SUM'
}
```

As you can see, the plan is to use this one place to configure aliases & aggregations as well.

edwinmp:
When `selectAll` is True, the `columns` property is used for ordering the results, at least for those columns that are provided. The rest are returned in whatever order.

edwinmp:
**Pending Features**

_SELECT_
- [x] Select All - tied to the `selectall` property. When provided, all table columns are returned. If the `columns` property is filled, this will be used to determine the order in which the provided columns will be returned. I don't think this bit will affect the SQL. I figure the fetched data will need to be processed before being returned by the API.

_JOIN_

- [x] We're adding a new property to the join config: `columns`. This property will indicate which columns of the joined table should be returned. It should match the `columns` property of the `select` config. I wonder if join table columns can be aggregated as well @wakibi

edwinmp:
Tested this and as long as the FE configs are valid, this will create the right SQL and return the expected data ... there's a potential bug for the select ordering when working with data sources that return many columns.

[...]

Issue #523, opened 2021-05-19
Advanced Query Builder API updates


edwinmp:
- [x] Return `advanced_config` in operation endpoints
- [x] Update `OperationSerializer` to handle the creation/update of operations with the `advanced_config`
- [x] Preview operations with the `advanced_config`
- [x] Return data for operations with the `advanced_config`
- [x] Generate aliases from the `advanced_config` (where applicable)

edwinmp:
Branch off `feature/advanced-query-builder`, which is the main branch for the Advanced QB. All PRs are made to this branch as well.

While I think this should still be handled in the already existing API code, all core code on the Advanced QB should be added under the `query_builder` app (guess this will be mostly functions & such) - unless you judge it otherwise. There's an `advanced.py` module for handling the actual `advanced_config` ... there's also a sample config at the bottom of that file. Making the `AdvancedQueryBuilder` class fully functional is beyond the scope of this issue so I'll create a different one for it. It actually seems like that functionality should come before this API update.

PR #553, opened 2021-06-23
Feature | Advanced QB API Updates

wakibi:
Issue #528 #523


edwinmp (commented):
Have fixed the tiny issues that were preventing testing ... can't really give it a properly thorough functional review without adding the FE bits that use it ... will branch off it and do that when I get time.

PR #555, opened 2021-06-24
Advanced query builder Select Aggregate

ThatcherK:


edwinmp:
Fixed this in the main select branch, so I'm closing this. Anything else left with the select like the validation shall be handled after some dependencies are settled.

Issue #562, opened 2021-07-08
Create, Update, Delete Advanced QB Datasets


edwinmp:
- [x] Create
- [ ] Update
- [x] Delete

**Create & Update Requirements**

- Name is required (should already be implemented i.e save button inactive otherwise)
- Description is required for non-draft datasets (should already be implemented i.e save button inactive otherwise)

**Update**

- Should populate config editors with existing JSON

PR #566, opened 2021-07-09
Feature | Advanced QB CRUD

ThatcherK:
Issue #562


edwinmp (commented):
Before you go even further, no redux ... I'm trying to get rid of it, so please don't create more components that depend on it.

PR #596, opened 2021-10-01
Add new columns to IATI spec

wakibi:


data_updates/Python/iati_transaction_spec.py
```diff
@@ -1051,14 +1060,17 @@ def flatten_activities(self, root):
                     x_transaction_value_usd = ""
                     x_reporting_org_type = recode_if_not_none(reporting_org_type_code, self.dictionaries["organisation_type"])
                     x_transaction_type = recode_if_not_none(transaction_type_code, self.dictionaries["transaction_type"])
+                    x_donor_transaction_type = recode_if_not_none(transaction_type_code, self.dictionaries["donor_transaction_type"])
```
akmiller01, line 1063: This might need an additional check before to avoid final value being `None`. Something like:
```
if transaction_type_code in self.dictionaries["donor_transaction_type"]:
    x_donor_transaction_type = recode_if_not_none(transaction_type_code, self.dictionaries["donor_transaction_type"])
```

data_updates/Python/iati_transaction_spec.py
```diff
@@ -1051,14 +1060,17 @@ def flatten_activities(self, root):
...
                     x_finance_type = recode_if_not_none(x_finance_type_code, self.dictionaries["finance_type"])
                     x_aid_type = recode_if_not_none(transaction_aid_type_code, self.dictionaries["aid_type"])
 
                     x_original_transaction_value = x_transaction_value
                     x_original_transaction_value_usd = ""
+                    x_original_transaction_value_USDm = ""
                     if transaction_convertable and x_currency:
                         if x_currency in self.dictionaries["ratedf"]:
                             x_original_transaction_value_usd = convert_usd(x_original_transaction_value, year, x_currency, self.dictionaries["ratedf"])
+                            x_original_transaction_value_USDm = (x_original_transaction_value_usd / 1000000) if x_original_transaction_value_usd and x_original_transaction_value_usd > 0 else x_original_transaction_value_usd
```
akmiller01: I think we can have negative values here, so we don't want to return the original negative value when it's less than zero. Would recommend replace with:
```
x_original_transaction_value_USDm = (x_original_transaction_value_usd / 1000000) if x_original_transaction_value_usd else ""
```

akmiller01 (changes requested):

akmiller01 (approved):

PR #606, opened 2021-11-11
Fix/ssl certs

wakibi:
This fixes SSL renewal issues.


akmiller01 (approved):
Looks like a good solution :+1:

PR #615, opened 2022-01-19
Feature | ETL Query

wakibi:
Adds ability to create a frozen dataset at the end of the ETL process. This is part of the Gates work. Once the ETL completes, this should kick in to create a "new" frozen dataset from the new updated IATI data.


wakibi:
## Instructions

1. Add ETLQuery via Django admin - This basically marks the query that will be run after the ETL process
2. Run the ETL process (for IATI in this case)
3. Access the data via {host}/api/savedqueryset/data/IATI/

I was supposed to run 1 and 2 over the weekend, but failed. So will trigger a manual simulation today.

edwinmp (approved):

Issue #557, opened 2021-06-28
Add handling for the HAVING clause on the Advanced QB


edwinmp:
**Resource**
https://www.postgresql.org/docs/13/sql-select.html
![image](https://user-images.githubusercontent.com/5672438/123625804-d0dea900-d818-11eb-9744-edb1fe153421.png)

**UI/Design**
- Similar if not identical to the filter UI

**Behaviour**
- Only works on columns that have an aggregation or are specified in the `groupby` clause
- When there's no aggregation or `groupby`, should show an error/warning when selected

PR #626, opened 2022-02-09
Feature | Advanced QB Having Clause

edwinmp:
Ticket #557


edwinmp:
@ThatcherK 
I created this PR so I can't really add code reviews via the files tab... anyway, this is a functional review.

- [x] Under the having conditions, I think we can have more than equal, less & greater ... maybe as many as we do for the filter
- [x] Change the `Replace` button caption to `Add` ... do this for the filter as well while you're at it.

For those that know what they're doing, this will be very useful ... great work.

ThatcherK:
@edwinmp done with the functional feedback

edwinmp:
@ThatcherK see the screenshot below - this is how the edit view appears for the having clause. It seems you're not doing a reverse mapping so the value is an object string.

image

2. When I try to groupby a string field & then I select the having clause, the app crashes with a massive error (probably cos it shouldn't be allowed)
3. Since your having conditions only work on numerical fields, `contains` should not be in the conditional list. They shouldn't only work on numerical fields or expressions, but I guess the most useful are those so lets leave them at that till it's requested.
4. I cannot add a having condition that isn't based on an aggregate ... e.g Amount USD > 20000 - somehow, the view forces the options on the value side to be aggregates. I guess this kind of usage would belong in a filter, but still, it's weird to force an aggregate.

frontend/src/components/JqueryQueryBuilder/JqueryQueryBuilder.tsx
```diff
@@ -32,9 +36,38 @@ const JqueryQueryBuilder: FunctionComponent<JqueryQueryBuilderProps> = ({
...
+            (previousValue as { column: string; aggregate: string }).aggregate,
+          ).map((option) => {
+            return {
+              value: `${previousValue?.column},${option}`,
+              label: `${option}(${previousValue?.column})`,
+            };
+          });
+        }
+      }
+      console.log(rules);
```
edwinmp: @ThatcherK Please do this outside of this component... it's what props are for cos this is crowding what was initially a simple component.

ThatcherK:
@edwinmp I need your review here

Issue #632, opened 2022-02-14
Force option for either Basic or Advanced Query Builder


edwinmp:
Currently, under the Query Builder sidebar item we have 2 options: Basic Query Builder and Advanced Query Builder. The Basic Query Builder is soon to be deprecated (not sure when), but before that happens, we have to drive up the usage of it's replacement. Simply asking people to use the Advanced QB hasn't yielded results, so we're going to force the choice.

- [x] Remove the multiple option dropdown for the Query Builder and simply have the Query Builder item.
- [x] On clicking Query Builder, show a modal that offers the option `Basic` or `Advanced` with the Advanced selected by default.
- [x] For each option selected in the modal above, show some info about each one (to be decided later - just show a description that makes sense for now)
- [x] Naturally, depending on the selected option, you get redirected to the matching QB

To not make things difficult for our users, we're going to need to give them the option to choose a default QB i.e. a `remember my choice` kind of thing.

- [x] Add a `preference` app & in its models.py file, add a `Preference` model that inherits from `BaseEntity`
- [x] Add the field `preferences` to the `Preference` model, a `JSONField` - give it the default value (from a variable) `{"queryBuilder": "advanced"}`
- [x] Add the field `global` to the `Preference` model  - a boolean field - give it a default value False - this naturally separates global preferences from user preferences. Make sure that in the Django admin, we can filter Preferences by this global field
- [x] Create an API endpoint for retrieving a user's preferences. Return both the global & user preferences (if any - default `{}`) - return object in format `{ global: {}, user: {} }` 
- [x] On the same view for retrieving, add one for updating preferences - naturally takes in a JSON object of the update preferences
- [x] On the FE, fetch these preferences on user login and save them in the local storage
- [x] When user clicks on Query Builder, check user preferences for preferred query builder - if none, show them the choice modal
- [x] Add a checkbox in the choice modal for `Remember my choice` - when checked, update user preferences with the selected option i.e `basic` or `advanced`

k8hughes:
@pauline-n can you update the ticket with some details about where you have got to and how long you think it will be before we can release

pauline-n:
Hi @k8hughes yes please I can. So far;

Under the Query Builder in the sidebar item, when clicked you get a page with a modal showing two options to choose from, with Advanced Query Builder showing by default.
Below is a picture showing how that is appearing:
[...]

PR #635, opened 2022-02-17
Feature | Force option for either Basic or Advanced Query Builder

pauline-n:
Feature | force option for either Basic or Advanced Query Builder. #632


wakibi:
Hi @pauline-n, please ensure you test (and by extension lint) your code well before marking a PR as ready. It will save all of us valuable time. Make sure it's passing the tests at the very least.

pauline-n:
Hi @wakibi , sorry about that. Working on that now then will re-request review.

Thanks
Pauline

ThatcherK:
@pauline-n Functional review

- [x] The advanced query builder is not selected by default
- [x] Am not comfortable with the labelling of the save button, could we label it along the lines of 'go to querybuilder' instead? For example when I don't check 'remember my choice', I won't know what to click next because I am not saving anything, yet the save button is the same that takes you to the query-builder. It's not very intuitive when it's labelled save.
- [x] On selection of a query builder of choice and checking the 'remember my choice' checkbox and saving, am taken to the desired querybuilder. But when I leave the querybuilder and and then try to access the querybuilder again via the sidebar menu, am given the modal pop -up again.I thought the choice is supposed to be saved. It looks like that isn't working

pauline-n:
@ThatcherK  thank you for the feedback, I must have missed out some steps let me get them working and will push commits afterwards

pauline-n:
Hello @ThatcherK have worked on the missing details above. Kindly review, all should be good now.

Thanks
Pauline

ThatcherK:
@ Pauline still on the functional side, 

- [x] When I select a query builder of choice, and check remember me, while still logged in, my choice is remembered, when I move away from the query builder , then later go back, am taken to the query builder I selected. Which is working perfectly well. The problem is when I log out and log back in, my choice is no longer remembered, I am given the modal pop-up again. 
But I have seen the reason , this is, check the code review to see why this is happening.

frontend/src/layouts/MainLayout.tsx
```diff
@@ -124,17 +125,18 @@ class MainLayout extends React.Component<MainLayoutProps, MainLayoutState> {
                 </Sidebar.Item>
                 <Sidebar.Item
                   active={this.state.activeRoute.includes('/queries/build/')}
-                  collapseId="query-builder"
+                  // collapseId="query-builder"
```
ThatcherK: Delete this commented out line, line 128

frontend/src/pages/Login/Login.tsx
```diff
@@ -107,13 +107,20 @@ export class Login extends React.Component<LoginProps, LoginState> {
...
         } else if (detail) {
           this.setState({ alert: detail });
         }
       })
       .catch(console.log); // tslint:disable-line
   };
 
+  private savePreference(token: string) {
+    getPreference(token).then((response) => response.json());
+    // .then(({ results }) => {
```
ThatcherK: Remove commented code, not being used

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+type SelectedQueryBuilder = 'basic' | 'advanced';
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+const QueryBuilderChooser: FC<RouteComponentProps> = (props: RouteComponentProps<{}>) => {
+  const [showModal, setShowModal] = useState(true);
+  const [selectedOption, setSelectedOption] = useState<SelectedQueryBuilder>('advanced');
+  const [checked, setChecked] = useState(false);
+  const [choice, setChoice] = useState<string>();
+  const [redirectPage, setRedirectPage] = useState('');
```
ThatcherK: This variable looks redundant, I don't see it's importance. Why did you add it? I feel like the variable selectedOption is enough, and the redirectPage variable is just a repetition of the same

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,128 @@
...
+          method: 'post',
+          url: api.routes.USERPREFERENCE,
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `token ${token}`,
+          },
+          data: userPreference,
+        })
+        .then(() => {
+          if (selectedOption === 'basic') {
```
ThatcherK: This if statement is redundant, not needed

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,128 @@
...
+          }
+        })
+        .catch((err) => console.log(err));
+    });
+    if (checked === true) {
+      localForage.setItem(localForageKeys.PREFERENCES, selectedOption);
+    }
+  };
+
+  if (redirectPage || choice) {
```
ThatcherK: Instead of redirectPage, use selectedOption

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,128 @@
...
+    if (redirectPage === 'basic' || choice === 'basic') {
```
ThatcherK: use selectedOption

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,128 @@
...
+        <Modal.Footer>
+          <CheckBox
+            label="Remember choice"
+            checked={checked}
+            onChange={() => {
+              handleChange();
+            }}
+          />
+          <Button
+            onClick={() => {
```
ThatcherK: Change this to `onClick={handleSave}`

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,128 @@
...
+          />
+          <Alert variant="secondary">
+            <p>The Advanced Query Builder</p>
+          </Alert>
+        </Modal.Body>
+        <Modal.Footer>
+          <CheckBox
+            label="Remember choice"
+            checked={checked}
+            onChange={() => {
```
ThatcherK: Change this to `onChange={handleChange}`

ThatcherK (changes requested):

edwinmp (changes requested):
Change the alert text for the basic option to:
`The original Query Builder. Creates queries using interconnected steps - no SQL knowledge required`

Change the text of the advanced option to:
`An improved UI with a lot more flexibility & options for creating advanced queries. Some of its features require a little SQL knowledge.`

Change the font size of the alert text to `14px`.

Why is the `Go to Query Builder` button purple? It's probably the ONLY purple button in the entire app ... Our colours are red, grey and white. For buttons, red & grey ... choose one. Classes are `btn-danger` or `btn-dark`. Also, Query Builder is 2 words.

Change from `Remember choice` to `Remember my choice`

Change title to `Select Query Builder`

Give the alert the class `p-2` to reduce on the padding so that it isn't too fat.

When a user clicks outside of the modal, instead of simply closing the modal and leaving them on a blank page, select & show the default QB (advanced)

core/urls.py
```diff
@@ -2,6 +2,7 @@
 from knox import views as knox_views
 from core.views import LoginView
 from core import views as core_views
+# from preference.views import PreferenceList, PreferenceDetail
```
edwinmp: Remove

core/urls.py
```diff
@@ -49,6 +50,9 @@
     path('dataset/alias/<int:pk>/', core_views.OperationColumnAlias.as_view()),
     path('dataset/estimate/<int:pk>/', core_views.EstimateQueryTime.as_view()),
     path('dataset/query/', core_views.GetOperationQuery.as_view()),
+    path('preferences/', core_views.PreferenceList.as_view()),
```
edwinmp: Remove - I don't see any use case where we'd fetch all the preferences... remove the view as well if it's doing nothing else. Wasn't part of the requirements in the first place.

core/urls.py
```diff
@@ -49,6 +50,9 @@
...
+    path('preferences/mine/', core_views.UserPreferenceList.as_view()),
+    path('preferences/<int:pk>/', core_views.PreferenceDetail.as_view()),
```
edwinmp: This too ... unless you're using this somewhere, I'd rather you just keep `preferences/mine` as it should be able to handle fetching and updating target user preferences.

core/serializers.py
```diff
@@ -373,16 +374,22 @@ class Meta:
...
+    user = serializers.ReadOnlyField(source='user.username')
+    class Meta:
+        model = Preference
+        fields = ('id', 'preferences', 'global_choice', 'user')
+
 class UserSerializer(serializers.ModelSerializer):
     tag_set = TagSerializer(many=True, read_only=True)
     operation_set = OperationSerializer(many=True, read_only=True)
     review_set = ReviewSerializer(many=True, read_only=True)
+    preferences_field = PreferenceSerializer(many=True, read_only=True)
```
edwinmp: I'm not sure what this is doing here - you're clearly not using it. Seems like a good idea to fetch and return the user preferences as part of the Login data, but it seems you abandoned that idea half way. It's up to you to pursue it - otherwise, remove this and the `fields` mention on like 392.

core/views.py
```diff
@@ -618,6 +623,34 @@ class UserDetail(generics.RetrieveAPIView):
...
+class UserPreferenceList(generics.ListCreateAPIView):
+    """ Returns a list of preferences for the current and authenticated user """
+    authentication_classes = [TokenAuthentication]
+    permission_classes = (permissions.IsAuthenticatedOrReadOnly & IsOwnerOrReadOnly,)
+    queryset = Preference.objects.all()
+    serializer_class = PreferenceSerializer
+
+    def perform_create(self, serializer):
+        serializer.save(user=self.request.user)
+    def get_queryset(self):
```
edwinmp, line 631: Blank line before this to separate the functions.

frontend/src/pages/Login/Login.tsx
```diff
@@ -107,13 +107,21 @@ export class Login extends React.Component<LoginProps, LoginState> {
...
         }
       })
       .catch(console.log); // tslint:disable-line
   };
 
+  private savePreference(token: string) {
+    getPreference(token)
+      .then((response) => response.json())
+      .then(({ results }) => {
+        localForage.setItem(localForageKeys.PREFERENCES, results[0].preferences);
```
edwinmp, line 121: You're diving right into `results[0].preferences`. This assumes there will always be results returned. It's safer to check.
```suggestion
        if (results.length) {
            localForage.setItem(localForageKeys.PREFERENCES, results[0].preferences);
        }
```

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
+import React, { FC, useState, useEffect } from 'react';
+import * as localForage from 'localforage';
+import { localForageKeys, api } from '../../utils';
+import { Alert, Button, Modal } from 'react-bootstrap';
+import { RouteComponentProps, useHistory } from 'react-router-dom';
+import { ICheckData, IRadio } from '../../components/IRadio';
+import { CheckBox } from '../../components/CheckBox';
+import axios from 'axios';
```
edwinmp: In your vscode for this file - `CTRL + SHIFT + P` -> then `Organise Imports`

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+type SelectedQueryBuilder = 'basic' | 'advanced';
+
+// eslint-disable-next-line @typescript-eslint/ban-types
```
edwinmp: Remove this, unless doing so results in an error ... I don't think so.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const toggleModal = () => {
+    setShowModal(!showModal);
+    history.push('/queries/build/advanced/');
+  };
+  const onRadioChange = (data: ICheckData) => {
+    setSelectedOption(data.value as SelectedQueryBuilder);
+  };
+
+  const history = useHistory();
```
edwinmp: This is a hook ... always have your hooks before all other code or you'll run into issues, so move this to where all the other hooks are.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+  useEffect(() => {
+    localForage.getItem<string>(localForageKeys.PREFERENCES).then((key) => {
+      if (key) {
+        setChoice(key || undefined);
+        if (key === 'advanced') {
+          history.push('/queries/build/advanced/');
+        } else {
+          history.push('/queries/build/basic/');
+        }
```
edwinmp:
```suggestion
        history.push(`/queries/build/${key}/`);
```

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+      }
+    });
+  }, []);
+
+  const handleChange = () => {
```
edwinmp: What's changing? With so many views, be as specific with your naming as possible. As u did on the radio listener, this can be `onCheckboxChange`

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { CheckBox } from '../../components/CheckBox';
+import axios from 'axios';
+
+type SelectedQueryBuilder = 'basic' | 'advanced';
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+const QueryBuilderChooser: FC<RouteComponentProps> = () => {
+  const [showModal, setShowModal] = useState(true);
+  const [selectedOption, setSelectedOption] = useState<SelectedQueryBuilder>('advanced');
+  const [checked, setChecked] = useState(false);
```
edwinmp, line 15: Give more descriptive name here ... `rememberChoice` & `setRememberChoice`

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const [choice, setChoice] = useState<string>();
```
edwinmp: You don't need both `selectedOption` & `choice` ... heck, they are practically synonyms ... make do with just one.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+          url: api.routes.USERPREFERENCE,
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `token ${token}`,
+          },
+          data: userPreference,
+        })
+        .then(() => {
+          if (selectedOption || choice) {
+            if (selectedOption === 'basic' || choice === 'basic') {
```
edwinmp: Clear evidence that you only need one of them.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+              history.push('/queries/build/basic/');
+            } else {
+              history.push('/queries/build/advanced/');
+            }
```
edwinmp: Simplify as I did somewhere up there... single line.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+          }
+        })
+        .catch((err) => console.log(err));
+    });
+    if (checked === true) {
+      localForage.setItem(localForageKeys.PREFERENCES, selectedOption);
```
edwinmp: Just realising that your endpoint is returning preferences as a single string instead of the JSON it should be... something is not right ... let me go back and check on that model.

edwinmp: The preferences are NOT supposed to only hold the option for the query builder ... which is why the instructions for this feature where very clear that the preferences should be a JSON field with a default value `{"queryBuilder": "advanced"}` ... that means under the `preferences` field, to get the query builder, you're supposed to parse the JSON and get the value of the `queryBuilder` property ... similary, that's what you update when saving a user preference. You do NOT save the query builder option directly into the preferences field. I hope you're following.

Why is it this way? Cos at some point, I'd like to add more preferences ... perhaps one that changes the UI theme to dark mode. I cannot do that if the preferences field has `basic` or `advanced` as the value. For this example, it'd probably be `{"queryBuilder": "advanced", "theme": "dark"}`.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+        </p>
+      );
+    } else {
+      return (
+        <p style={{ fontSize: 14 }}>
+          An improved UI with a lot more flexibility & options for creating advanced queries. Some
+          of its features require a little SQL knowledge.
+        </p>
+      );
+    }
```
edwinmp: Rewrite with a single return statement.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
```diff
@@ -0,0 +1,134 @@
...
+            inline
+            checked={selectedOption === 'advanced'}
+          />
+          <Alert variant="secondary" className="p-2">
+            {alertMessage()}
+          </Alert>
+        </Modal.Body>
+        <Modal.Footer>
+          <CheckBox label="Remember my choice" checked={checked} onChange={handleChange} />
+          <Button onClick={handleSave} variant="btn-danger" className="btn-danger">
```
edwinmp, line 105: Ah, I wondered why every time I clicked `Go to Query Builder` a new preference was added to the database ... my user now has like 20 preferences. Clicking this button shouldn't automatically save. It should only save if a user has chosen  `Remember my choice`. Therefore, I suggest you create a different function for this button that automatically navigates to the user's choice without saving IF they haven't chosen to remember their choice. Else, save.

frontend/src/utils/api.ts
```diff
@@ -27,5 +27,7 @@ export const api = {
     DOWNLOAD_SAVED_QUERYSET: '/api/tables/download/{table}/dataset/',
     FETCH_OPERATION: '/api/operations/{id}',
     DATASET_QUERY: '/api/dataset/query/',
+    PREFERENCES: '/api/preferences/',
```
edwinmp: Remove this ... not used.

frontend/src/utils/index.ts
```diff
@@ -178,3 +179,13 @@ export const getColumnFromName = (
...
   columns.find((col) => col.name === name) as AdvancedQueryColumn;
+
+export const getPreference = (token: string) => {
+  return window.fetch(api.routes.USERPREFERENCE, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `token ${token}`,
+    },
+  });
```
edwinmp: Would be best if you returned the actual JSON here i.e `response.json()` ... abstract as much within this function as you can so that usages of it don't have to do as much ... also, lets see you can do it with `async ... await`.

preference/models.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+from core.models import BaseEntity
+
+# Create your models here.
+def default_value():
+    return {"queryBuilder" : "advanced"}
+class Preference(BaseEntity):
```
edwinmp, line 8: Space before ... separate functions, classes ... any of those kinds of big blocks of code. It makes for cleaner code.

edwinmp (changes requested):
The bug is gone but there is still some work to do.

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
pauline-n: Reason why i have both is so that someone can still be redirected to the querybuilder without the querybuilder of their choice being saved. Then should they select the remember choice checkbox, their choice will be saved and also they will be redirected. However, Let me do away with one so that they have no option but to select the remember choice checkbox.

edwinmp: Nope... they don't need to be forced to select. You can do this with only one of them without changing the existing behaviour. Also, if a change I've suggested requires you to change the behaviour, DON'T do it till you check with me.

pauline-n: Noted. I have some that are doing so. I will check in with you in the morning then

frontend/src/pages/QueryBuilderChooser/QueryBuilderChooser.tsx
pauline-n: Am not sure I have understood this well,

edwinmp: I hope you now understand what I mean.

Issue #627, opened 2022-02-09
Update Advanced QB Filter options & minor behaviour


edwinmp:
**Behaviour Updates**

- [x] Sort columns alphabetically

**What's Missing**

- [x] Not Equal
- [x] Less Than or Equal
- [x] Great Than or Equal
- [x] E.T.C - Basically, it can't do everything the current basic filter can do.

**Resources**
https://querybuilder.js.org/

_Originally posted by @edwinmp in https://github.com/devinit/ddw-analyst-ui/issues/531#issuecomment-963890218_

PR #636, opened 2022-02-21
Feature | Add Missing Filter Operators

davidebukali:
Add comparison operators #627


edwinmp (approved):
Looks good but still have to test it on dev, but want to keep the charts running so I'll have to add it to a release branch first.

PR #640, opened 2022-02-24
Enable spotlights development in dev compose file

wakibi:
@edwinmp check this to enable testing spotlights in analyst UI. See the comment I left on line 23 in `docker-compose.dev.yml` file. Also, just run `docker-compose -f docker-compose.dev.yml up spotlights` in case you are testing spotlights strictly. That should save you some resources

### DIFF
--- a/Dockerfile_dev_pg
+++ b/Dockerfile_dev_pg
@@ -1,0 +1,11 @@
+# start with a base image
+FROM postgres:13.2
+LABEL maintainer="akmiller01 <Alex Miller, alex.miller@devinit.org>"
+
+# New users
+ENV POSTGRES_USER=analyst_ui_user
+ENV POSTGRES_PASSWORD=analyst_ui_pass
+
+
+# Add initialization script to initdb
+ADD ./database_setup/dba.sql /docker-entrypoint-initdb.d/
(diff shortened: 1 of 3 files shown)


Dockerfile_dev_pg
```diff
@@ -0,0 +1,11 @@
+# start with a base image
+FROM postgres:13.2
+LABEL maintainer="akmiller01 <Alex Miller, alex.miller@devinit.org>"
```
edwinmp: Why are u giving away your credit? :D

docker-compose.dev.yml
```diff
@@ -4,9 +4,29 @@ services:
...
+
+  spotlights:
+    build:
+      context: .
+      dockerfile: Dockerfile_spotlights_dev
+    expose:
+      - "8000"
+    # Replace below path with the path to your ddw-external-api repo
+    volumes:
+      - "/Users/boss/chris/git/ddw-external-api:/src/ddw-external-api"
```
edwinmp: I think this would work best here https://docs.docker.com/compose/environment-variables/

edwinmp (changes requested):

edwinmp (approved):

Issue #591, opened 2021-09-23
Advanced Query Builder - Adding Multiple JOINs


JennyR-DI:
I've been trying to add ISO code information to both the Donor and Recipient columns in the 'OECD DAC2B: OOFs and export credits' dataset.
I can add the first LEFT JOIN no problem:
![image](https://user-images.githubusercontent.com/90774534/134512886-5fb397d9-2487-492a-9c3c-1edb2a054d50.png)

But I can't add a second LEFT JOIN to obtain the ISO values for the 'Recipient Code' column.  Can you please advise how I can add two JOINS in a query?

I've tried saving the query after the first JOIN and then editing it to add the second JOIN but this hasn't worked for me so far.

edwinmp:
@JennyR-DI we don't yet have support for multiple joins ... I'll mark this ticket as a feature request.

JennyR-DI:
Thanks Edwin for clarifying that, that's good to know I shouldn't do multiple joins now.

wakibi:
Adding for emphasis so we do not get, this includes multiple self joins i.e same table being joined more than once

k8hughes:
@davidebukali  and @edwinmp can you tag @JennyR-DI or @dean-breed when this is on staging for testing

davidebukali:
@JennyR-DI this is ready for testing on [staging](https://staging-ddw.devinit.org/)

PR #644, opened 2022-03-09
Feature | Multiple Joins

davidebukali:
Add multiple joins to the advanced query builder #591

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -187,6 +187,7 @@ jobs:
           npm ci
           npm run build
           docker-compose exec -T web python manage.py collectstatic --no-input
+          docker-compose restart web
 
   end-to-end:
     name: 'Cypress End-to-End'
--- a/deploy.sh
+++ b/deploy.sh
@@ -50,6 +50,7 @@ echo "Generating Javascript Assets"
 npm i
 npm run build
 docker-compose exec -T web python manage.py collectstatic --no-input
+docker-compose restart web
 
 echo "Fetch CSV Files"
 
--- a/deploy.sh
+++ b/deploy.sh
@@ -26,9 +26,14 @@ cp certbot_success.sh ./ssl
 chmod +x ssl/certbot_success.sh
 mkdir -p certbot_logs
 
+echo "Install JS dendencies"
+
+nvm install 16.14.2
+nvm use 16.14.2
+npm i
+
 echo "Build JS"
 
-npm i
 npm run build
 
 echo "Building docker"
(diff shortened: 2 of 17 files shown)


edwinmp (changes requested):
There's more than a few behavioural bugs. Test better. Proper testing is NOT just using it as expected ... users don't behave that way.

Here's the issues I found

1. I have to click `Add` or `Edit` to update the join config - I found this a UX hustle. I press the plus button, it tells me the mapping has been added - that's all I need.
2. It's not clear how to add multiple mappings for the same table - I figured it out by mistake. So like the basic join, use multiple rows for each new mapping - I don't mind if you have to reuse that view, complete with the whole `Add mapping` button ... it's clearer & our users are already used to it.
3. The select columns widget is buggy ... for one, it seems to be tied to its twin under the `Select` query. So when the select query is set to `Select all`, the one in the join has `Select columns for Ordering`. Order columns hasn't worked once ... if it's not disabled, it just doesn't have the expected columns (I think it's taking from the `Select` config). Clear/reset does nothing as far as I can tell. If you want to customise this component for reuse, be sure it's fully customisable ... e.g besides all the above, it'd be nice to have a custom button caption like `Select Join Columns` or something. Also, joins don't require columns from the joined table to work so not having any selected shouldn't keep the `Add` button disabled. A mapping is all that's required.
4. Delete a join and then try to add a new one ... fails.
5. There's a strange delay to update the preview (any of them), For the config preview, I have to click on it for the changes to show. I'm not sure this is only on your join or caused by it.
6. Remove the icon on the `View joins` buttons
7. It's confusing having actions all over the place ... like the `Select`, I'd rather they're all in a single ButtonGroup (including the mapping & delete) - not important for now, so don't mind it - can improve later (behaviour first).
8. Can't seen to add more than one join ... the `Add` button never seems to light up the second time around. And then it magically lit up when I scrolled through the configs (maybe I clicked the editor)

wakibi:
@davidebukali is this ready for review? I see you pushed more changes after the last changes were requested

davidebukali:
@wakibi its not ready for review, am working on the feedback

davidebukali:
@edwinmp and @wakibi this includes feedback for issues 1,2,4 and 8 mentioned in the comment above; From the pending issues, I think 3 would  require most work, while the rest are more likely simple edits.

edwinmp:
@davidebukali create a beta release so this can be tested on staging

edwinmp:
Since yours is the incoming PR, could you please generate migrations & commit - there's a package update that requires it.

edwinmp:
@davidebukali could you please sort the failing e2e test as well ... it's just the one so it's probably nothing serious.

edwinmp (approved):

PR #645, opened 2022-03-10
Feature | Advanced QB Having Clause

ThatcherK:
ticket #557


frontend/src/components/AdvancedHavingQueryBuilder/AdvancedHavingQueryBuilder.tsx
```diff
@@ -0,0 +1,142 @@
...
+            onClick={onReset}
+          >
+            Reset
+          </Button>
+        </ButtonGroup>
+      </>
+    );
+  }
+
+  return <Alert variant="warning">Having clause requires numeric columns or aggregates</Alert>;
```
edwinmp: As far as I know, it only requires the `group by` clause - if your implementation only supports numeric comparison, I get that, but right now I'm getting this error when I don't have a `group by` clause. I think at the very least, it should check for that requirement - both if the numeric is all you support.

frontend/src/components/JqueryQueryBuilder/JqueryQueryBuilder.tsx
```diff
@@ -4,22 +4,27 @@ import React, { FunctionComponent, useEffect } from 'react';
...
 
 export interface JqueryQueryBuilderProps {
   fieldData: JqueryQueryBuilderFieldData[];
   getJqueryBuilderInstance: (jqInstance: any) => void;
   icons?: JqueryQueryBuilderIcons;
-  rules: JqueryQueryBuilderFilter;
+  rules: JqueryQueryBuilderFilter | JqueryQueryBuilderHaving;
+  getHavingQueryValues?: (
+    rules: JqueryQueryBuilderFilter | JqueryQueryBuilderHaving,
+  ) => JqueryQueryBuilderFilter | JqueryQueryBuilderHaving;
```
edwinmp, line 19: The fact that you need to make updates to this component tells me it wasn't built right ... oh well, will create a ticket to move us to the ReactQueryBuilder.

frontend/src/components/JqueryQueryBuilder/utils/index.tsx
```diff
@@ -35,6 +37,73 @@ export const parseQuery = (
...
+            });
+          }
+        }
+      }
+    }
+  }
+
+  return finalElement;
+};
+
```
edwinmp: This definitely shouldn't be here since the `JqueryQueryBuilder` shouldn't care what you're using it for. Since this is dealing with the having clause, it should be in the utils of the `AdvancedHavingQueryBuilder`. Applies to filter code if it was placed in here.

frontend/src/utils/hooks/operations.ts
```diff
@@ -44,13 +45,15 @@ const PREVIEWBASEURL = api.routes.PREVIEW_SINGLE_DATASET;
...
+      return { status, error: data.results[0].error as string };
+    }
+
     return { data: data.results, status };
   } else if (status === 401) {
     setToken('');
 
     return { status, error: 'invalid token' };
-  } else if (data.results && data.results.length && data.results[0].error) {
-    return { status, error: data.results[0].error as string };
```
edwinmp: I understand copying it, but are you sure this can't happen? I'm sure this was meant to handle `any other result` that's not status 200, 201 or 401 ... so I think you'd best be safe and leave this.

query_builder/advanced.py
```diff
@@ -122,8 +123,16 @@ def handle_join(self, table, query, config):
...
+            if config.get('selectall'):
+                raise LookupError('Columns must be explicitly SELECTED for queries that use GROUP BY clauses')
+            if 'columns' in config:
+                query = self.get_groupby_query(table, query, config.get('groupby'), config.get('columns'))
+                if 'having' in config:
+                    query = self.get_having_query(table, query, config.get('having'))
+            else:
+                raise LookupError('Columns must be explicitly SELECTED for queries that use GROUP BY clauses')
+
+        if 'having' in config and 'groupby' not in config:
```
edwinmp: This should not happen! If it does, return an error ... `HAVING` is to `GROUP BY` as `WHERE` is to `SELECT` ... you can't have one without the other.
I don't think it should appear in this function at all ... it should be in the `get_groupby_query` function

query_builder/advanced.py
```diff
@@ -122,8 +123,16 @@ def handle_join(self, table, query, config):
 
     def get_select_query(self, table, query, config):
         if 'groupby' in config:
-            query = self.get_groupby_query(table, query, config.get('groupby'))
-        if 'having' in config:
+            if config.get('selectall'):
```
edwinmp: This is a good check, but in the wrong position ... it should come first within this function because when true, it should result in error for not just the `having` config but the `groupby` as well. So I see:

```
if config.get('selectall'):
    # validate that groupby and having configs are not provided - if so, raise an error - 
    # else, generate selectall query & return
if 'columns' in config:
    # handle `groupby` and `having` configs & the rest here, in that order
else:
    # raise an error
   
```

query_builder/advanced.py
```diff
@@ -156,9 +165,20 @@ def get_select_query(self, table, query, config):
 
         return query
 
-    def get_groupby_query(self, table, query, columns):
-        # TODO: handle .having here as its usage is based on the groupby
```
edwinmp, line 160: I see I mentioned it here

query_builder/advanced.py
```diff
@@ -176,11 +196,13 @@ def get_filter_query(self, table, query, config):
...
             rootConfig = config.get(self.andKey)
+            # a way to handle complex having configs
+            crit = Criterion.all([ self.get_having_criterion(table, config) for config in rootConfig ])
         elif self.orKey in config:
             rootConfig = config.get(self.orKey)
+            # a way to handle complex having configs
+            crit = Criterion.all([ self.get_having_criterion(table, config) for config in rootConfig ])
 
-        # a way to handle complex having configs
-        crit = Criterion.all([ self.get_having_criterion(table, config) for config in rootConfig ])
```
edwinmp, line 183: Why did u duplicate this on line 199 and 203 when having it here is exactly the same thing, only without the duplication? Or do you expect a scenario where there's neither AND/OR? hmmm... okay - handle it properly then.

edwinmp (changes requested):

query_builder/advanced.py
ThatcherK: Thinking about it, this check is not needed even in the `get_groupby_query` function, because the frontend won't allow a having clause without a `groupby`. And by the `get_groupby_query` function is called only when there is a `groupby` in the config

ThatcherK:
@edwinmp done with the feedback

query_builder/advanced.py
edwinmp: This is what you were supposed to change, instead you changed the `get_filter_query`. Please revert that ... as for this bit, leave it as it but change the Criterion on line 203 from `Criterion.all` to `Criterion.any` ...

query_builder/advanced.py
```diff
@@ -137,50 +143,65 @@ def get_select_query(self, table, query, config):
...
 
     def get_having_query(self, table, query, config):
         if self.andKey in config:
             rootConfig = config.get(self.andKey)
+            # a way to handle complex having configs
+            crit = Criterion.all([ self.get_having_criterion(table, config) for config in rootConfig ])
         elif self.orKey in config:
             rootConfig = config.get(self.orKey)
+            # a way to handle complex having configs
+            crit = Criterion.all([ self.get_having_criterion(table, config) for config in rootConfig ])
```
edwinmp: `Criterion.any`

query_builder/advanced.py
```diff
@@ -137,50 +143,65 @@ def get_select_query(self, table, query, config):
...
             rootConfig = config.get(self.andKey)
-            # a way to handle complex filter configs
-            crit = Criterion.all([ self.get_filter_criterion(table, config) for config in rootConfig ])
         elif self.orKey in config:
             rootConfig = config.get(self.orKey)
-            # a way to handle complex filter configs
-            crit = Criterion.any([ self.get_filter_criterion(table, config) for config in rootConfig ])
+
+         # a way to handle complex filter configs
+        crit = Criterion.any([ self.get_filter_criterion(table, config) for config in rootConfig ])
```
edwinmp: This is a breaking change. Please revert.

edwinmp (changes requested):
A little mistake there ...

ThatcherK:
@edwinmp done

edwinmp (approved):

Issue #647, opened 2022-03-14
Create data update pipeline for the Gates dashboard


edwinmp:
Currently, the CSV files are generated manually for OECD, and the IATI pipeline still has some manual elements. Dean has already done most of the work for the IATI in PR #628  & we just have to do the same for OECD.

edwinmp:
A fully configured component is already available in [this](https://github.com/devinit/ddw-analyst-ui/tree/feature/having-clause) branch

k8hughes:
@ThatcherK is this nearly ready to release?  I think IATI is done and OECD is still to do - is that right?

k8hughes:
@dean-breed can you also follow up this one so we know the dashboard is updating

dean-breed:
We need the [shell file](https://github.com/devinit/ddw-analyst-ui/blob/feature/iati-gates/data_updates/iati_transactions.sh) to be pulled into master if everyone is happy

ThatcherK:
I am going to make a few more edits for the csv file names, make releases and test if the changes work well without breaking the charts.

dean-breed:
Great, thanks @ThatcherK

ThatcherK:
This is done, the files have been tested on the live site, nothing is broken

dean-breed:
Thanks @ThatcherK

ThatcherK:
@dean-breed , hadn't yet been merged

PR #653, opened 2022-03-15
Feature/OECD Data Pipeline

ThatcherK:
ticket #647 
The csvs generated from this have been pushed to gates/oecd branch on the di-website-data repo


ThatcherK:
@dean-breed Please take a look at this data update script for the gates OECD csv data

dean-breed (dismissed):
This all looks good. Thanks @ThatcherK !

dean-breed:
@ThatcherK I will now be working on creating the final dataset 'donor-by-recip-2019.csv' which I will add to this script to push to gh too

edwinmp (changes requested):
Why do you have the CSV files in here? If those are the files generated by running this script, then I'd rather they don't get committed - add the entire folder to `.gitignore`

Won't merge this in until @dean-breed's pending work on it is done.

dean-breed:
I have added the new CSV code @edwinmp

dean-breed:
But I need to change one thing... Wait up @edwinmp

dean-breed:
Done @edwinmp

data_updates/Python/oecd_data.py
```diff
@@ -0,0 +1,124 @@
...
+import numpy
+from numpy import NaN
+import pandas as pd
+from github import Github, InputGitTreeElement
+from django.conf import settings
+
+PURPOSE_CODE_TRENDS_URL = "https://staging-ddw.devinit.org/api/export/1241/"
+CSV_FOLDER = "oecd_csv"
+DATA_REPO = "devinit/di-website-data"
+REMOTE_BRANCH = "gates/oecd"
```
edwinmp: Test the generated files on the `di-chart-boilerplate` repo (`page/gates` branch) to make sure the tables and charts still function, then change this to `main`.

edwinmp (changes requested):
@ThatcherK still seeing the CSV files. Once they're committed, you have to delete them and commit before that `.gitignore` change takes effect.

edwinmp:
Lastly, this isn't plugged into any pre-existing update process so it won't deploy updates automatically ... there's an `oecd.sh` but I'm not sure if that's part of a regular update pipeline ... @wakibi do we have a scheduled event or such script for auto updating OECD data?

ThatcherK:
@dean-breed , i was comparing the recipient by donor table form the previous csv, with the current and noticed a slight difference, for united states, reproductive health care and family planning. Previously the first donor was `Unspecified` , its now `Bilateral, unspecified`. Is this the desired result?

dean-breed:
@ThatcherK Thanks for pointing this out. I think it's fine but Mattie and Duncan are taking a look today and I'll make sure they agree!

dean-breed:
@wakibi Good to go, I think!

wakibi:
> Lastly, this isn't plugged into any pre-existing update process so it won't deploy updates automatically ... there's an `oecd.sh` but I'm not sure if that's part of a regular update pipeline ... @wakibi do we have a scheduled event or such script for auto updating OECD data?

@edwinmp not one I know of. @dean-breed, is this the intended functionality? A one-off or manual update not tied to any ETL process?

dean-breed:
@wakibi We can just trigger this script when the DAC data is updated

edwinmp:
> @wakibi We can just trigger this script when the DAC data is updated

@dean-breed @ThatcherK  then you're going to need to create a `.sh` script that we can configure to trigger manually on via the scheduled events feature ... similar to [this](https://github.com/devinit/ddw-analyst-ui/blob/develop/data_updates/iati_transactions.sh)

dean-breed:
@ThatcherK are you okay to do this?

ThatcherK:
@dean-breed  sure I'll look into it

data_updates/Python/oecd_data.py
```diff
@@ -0,0 +1,127 @@
...
+
+# RH FP channels OECD
+
+print("Starting read-in: Channels")
+
+channels_data = pd.read_csv(ODA_CHANNEL_TYPE_URL)
+channels_data = pd.DataFrame(channels_data)
+channels_data.columns = ["year","donor_name","purpose_code","purpose_name","oecd_channel_parent_name","oecd_aggregated_channel","usd_disbursement_deflated_Sum"]
+
+channels_data.to_csv(f'{CSV_FOLDER}/RH FP channels OECD.csv', encoding='utf-8', index=False)
```
wakibi: @ThatcherK sorry this is coming late, but would be nice to have the folder 'oecd_csv' created on the fly if it does not exist already. Otherwise everything will crush (or at the very least handle it gracefully). Also, can we have the file names as one word? Use underscores ("_") in case you really want to keep it descriptive. Sometimes filenames with spaces can behave unexpectedly.

data_updates/Python/oecd_data.py
```diff
@@ -0,0 +1,127 @@
...
+
+# OECD purpose code trends data
+
+print("Starting read-in: Purpose")
+
+purporse_code_data = pd.read_csv(PURPOSE_CODE_TRENDS_URL)
+purporse_code_data = pd.DataFrame(purporse_code_data)
+purporse_code_data.columns = ["year","donor_name","purpose_code","purpose_name","usd_disbursement_deflated_Sum"]
+
+purporse_code_data.to_csv(f'{CSV_FOLDER}/RH and FP Purpose code trends chart OECD.csv', encoding='utf-8', index=False)
```
wakibi: Same as above

data_updates/Python/oecd_data.py
```diff
@@ -0,0 +1,127 @@
...
+
+# RH FP aid type OECD
+
+print("Starting read-in: Aid type")
+
+aid_type_data = pd.read_csv(ODA_AID_TYPE_URL)
+aid_type_data = pd.DataFrame(aid_type_data)
+aid_type_data.columns = ["donor_name","aid_type_di_name","year","purpose_name","purpose_code","usd_disbursement_deflated_Sum"]
+
+aid_type_data.to_csv(f'{CSV_FOLDER}/RH FP aid type OECD.csv', encoding='utf-8', index=False)
```
wakibi: Same as above

data_updates/Python/oecd_data.py
dean-breed: I'll leave @ThatcherK to do the folder creation but wanted to say that the only reason they are the file names is because that's what DK called them originally. I agree that underscores are better

data_updates/Python/oecd_data.py
```diff
@@ -0,0 +1,127 @@
...
+from django.conf import settings
+
+PURPOSE_CODE_TRENDS_URL = "https://staging-ddw.devinit.org/api/export/1241/"
+CSV_FOLDER = "oecd_csv"
+DATA_REPO = "devinit/di-website-data"
+REMOTE_BRANCH = "main"
+REMOTE_FOLDER = "2022"
+ODA_AID_TYPE_URL = "https://staging-ddw.devinit.org/api/export/1238/"
+ODA_CHANNEL_TYPE_URL = "https://staging-ddw.devinit.org/api/export/1237/"
+ODA_RECIP_TYPE_URL = "https://ddw.devinit.org/api/export/1339/"
```
wakibi: For the constants here that may need changing... Can we find a better way to handle them? You have hard-coded the URLs for example and they are staging URLs, are they the same ones that will be used on production? The Remote folder shows 2022, shall we maintain 2022 next year as well or we shall update it to 2023? If the year will keep changing, I propose using the datetime functions to get the current year. For the URLs, possibly we use Environment variables and fall back to the staging URLs if no environment variable has been found.

wakibi (changes requested):

ThatcherK:
@wakibi @edwinmp , all changes made. I now have to change csv names on the gates repo, to match the ones with underscores...

ThatcherK:
@wakibi  and @edwinmp , please can we find an agreed format for the naming of the csv files, originally they were space separated words, Chris suggested underscore separated. What would be preferable?

edwinmp:
@edwinmp go with whatever Chris has recommended ... just be sure to update everything that will be affected by that change (several places) ... get in a huddle and I'll talk you through it.

ThatcherK:
@wakibi , should this be merged?

wakibi (dismissed):
LGTM

edwinmp (dismissed):

data_updates/Python/oecd_data.py
```diff
@@ -0,0 +1,137 @@
...
+import os
+import glob
+import numpy
+from numpy import NaN
+import pandas as pd
+from github import Github, InputGitTreeElement
+from django.conf import settings
+
+current_date = datetime.datetime.now()
+PURPOSE_CODE_TRENDS_URL = os.getenv('PURPOSE_CODE_TRENDS_URL') or "https://staging-ddw.devinit.org/api/export/1241/"
```
edwinmp: I think the default can be added directly in the `getenv` function i.e
`os.getenv('PURPOSE_CODE_TRENDS_URL', 'https://staging-ddw.devinit.org/api/export/1241/')`

... do for the others as wells

edwinmp (dismissed):

ThatcherK:
Tested on staging, everything looks good now!! I need an approving review to merge

edwinmp (approved):

Issue #652, opened 2022-03-15
Dataset sources don't show for AQB datasets


edwinmp:
![image](https://user-images.githubusercontent.com/5672438/158372839-92742fec-1c7d-40fa-b113-c230a54deb55.png)

edwinmp:
Will be deployed in the next release

Issue #651, opened 2022-03-15
Dataset filter by data source doesn't return AQB datasets


edwinmp:
**Steps to Reproduce**

1. Click on `My Datasets` on the sidebar
2. Ensure you have datasets created via the Advanced Query Builder
3. Click `Filter by data source` and select a matching data source to one on an AQB query
4. See bug i.e. Query will not show up in results

edwinmp:
Will be deployed in the next release.

k8hughes:
@edwinmp will it be possible to get this on staging in the next release?  Might be good to get a release out soon so that we don't get merg conflicts.

edwinmp:
Merge conflicts are rare and easy to deal with @k8hughes ... no worries - will deploy all this tomorrow.

PR #654, opened 2022-03-16
Fix | Dataset Listing for AQB Datasets

edwinmp:
Ticket #652 #651


ThatcherK (approved):
Looks okay works fine

PR #692, opened 2022-05-13
Frozen data sets split from source data in the data sources view

### LINKED ISSUE #687: Frozen data sets split from source data in the data sources view
In the data source menu there two types of data sets also need to be split out.  So as a user when I go to browser the data sets in the DDW I can select if I want to see source data or frozen data then, the page returned will show only one of these types.

https://github.com/devinit/ddw-analyst-ui/issues/663

wakibi:


davidebukali (commented):
LGTM!

edwinmp (dismissed):
Looks good ... Deploy a beta to staging to test

edwinmp:
@wakibi Your super long branch names won't likely catch on ... I don't think branch names need to be sentences to make sense ... if u want details, read the PR title/description hehe

frontend/src/components/SourcesTableCard/SourcesTableCard.tsx
```diff
@@ -33,29 +34,39 @@ export const SourcesTableCard: FunctionComponent<SourcesTableCardProps> = (props
...
     const queryParameters = { ...queryString.parse(search), ...filter };
     const cleanParameters: Partial<TableFilters> = {};
     if (queryParameters.search) cleanParameters.search = queryParameters.search;
     if (queryParameters.page) cleanParameters.page = queryParameters.page;
+    if (queryParameters.frozen) cleanParameters.frozen = queryParameters.frozen;
     history.push(`${pathname}?${queryString.stringify(cleanParameters)}`);
   };
 
+  const dropDownValues = [
+    { key: 'non-frozen', text: 'Non-frozen Sources', value: 0 },
```
edwinmp: Rename this `Core Sources` or something that's not `non-frozen` :)

edwinmp (changes requested):
Don't think the BE changes where worth it here ... frozen data sources are too few (and soon to be policed) that a FE filter would have sufficed, as we'd agreed. Would have been a lot less code.

wakibi:
> 

That's correct, and was my plan originally. I just thought it would give the wrong impression on the paginator when I started implementing, since the sources are fetched in batches of 20, and the count is got from the API as well. It would have made sense if we were fetching all sources at once. Your thoughts?

edwinmp:
> > 
> 
> That's correct, and was my plan originally. I just thought it would give the wrong impression on the paginator when I started implementing, since the sources are fetched in batches of 20, and the count is got from the API as well. It would have made sense if we were fetching all sources at once. Your thoughts?

Oh, I'd not accounted for the `paginator` ... you're absolutely right ... let's keep it as is.

edwinmp (approved):

Issue #663, opened 2022-04-11
Split out frozen data sets from source data


dean-breed:
Split frozen datasets away from the source datasets in the selection menus. Create two separate options and/or put frozen datasets at the bottom of the selection menu, whichever is easiest.

k8hughes:
@wakibi can you see if this breaks into two tickets - or if one can you add some more details so it is clear for whoever does the work?

wakibi:
@dean-breed would you still want the frozen datasets to be queried? Or only downloaded?

k8hughes:
In the query builder there needs to be a check box where a user can select which drop down they see.
Options:
* Source data sets
*  Frozen data sets
Once a user selects from these options the appropriate drop down menu will be available.

In the data source menu there two types of data sets also need to be split out.  So as a user when I go to browser the data sets in the DDW I can select if I want to see source data or frozen data then, the page returned will show only one of these types. - This work is in issue https://github.com/devinit/ddw-analyst-ui/issues/687 

This is the issue for the query builder.

dean-breed:
@wakibi We briefly discussed this over slack but the issue in #687 is exactly what I mean, yes.

davidebukali:
@dean-breed this is ready for testing on [staging](https://staging-ddw.devinit.org/)

dean-breed:
I've had a play around with this and it looks good to me! Happy to go ahead and deploy live the next time we do @davidebukali

davidebukali:
Thank you @dean-breed

PR #693, opened 2022-05-16
Feature | Split Frozen and Normal Datasources

davidebukali:
#663 

Basic QB
Screenshot 2022-05-23 at 12 22 35

Advanced QB
Screenshot 2022-05-23 at 12 23 23


frontend/src/utils/hooks/sources.ts
```diff
@@ -34,7 +34,7 @@ export const useSources = (options: Options = defaultOptions, fetch = false): Li
   const fetchSources = (): void => {
     const url = `${api.routes.SOURCES}?limit=${options.limit}&offset=${options.offset}&search=${
       options.search || ''
-    }`;
+    }&frozen=2`;
```
edwinmp: I saw a similar change on Chris' branch, so I expect a clash ... so either branch off his, or wait in line to resolve the conflict ... frankly, I'd have preferred if you used the API endpoint to fetch only exactly what you need, but I also don't think an API change was worth it in this case given how few frozen data sources we have & how we're planning on clearing them out every few months. So, go ahead with the filter but consider how Chris' changes work to eliminate the conflict.

edwinmp (changes requested):
Will review more once the functional change has been made.

frontend/src/utils/hooks/sources.ts
davidebukali: Rebased `develop`, now using Chris' version

frontend/src/components/DataSourceSelector/DataSourceSelector.tsx
```diff
@@ -1,20 +1,31 @@
...
 
 const DataSourceSelector: FunctionComponent<ComponentProps> = ({ source, ...props }) => {
   const { sources } = useContext(SourcesContext);
+  const [selectedDataSource, setSelectedDataSource] = useState<List<SourceMap>>(List());
+  useEffect(() => {
+    if (props.datasourceType === 'non-frozen') {
+      setSelectedDataSource(sources.filter((item) => item.get('schema') === 'repo'));
+    } else {
+      setSelectedDataSource(sources.filter((item) => item.get('schema') !== 'repo'));
+    }
```
edwinmp:
```suggestion
    setSelectedDataSource(sources.filter((item) => props.datasourceType === 'non-frozen' ? item.get('schema') === 'repo' : item.get('schema') !== 'repo'));
```

frontend/src/components/DataSourceSelector/DataSourceSelector.tsx
```diff
@@ -24,16 +35,16 @@ const DataSourceSelector: FunctionComponent<ComponentProps> = ({ source, ...prop
...
+      <label>Select {props.label}</label>
       <Dropdown
         placeholder="Select Data Source"
         fluid
         selection
         search
-        options={getSelectOptionsFromSources(sources)}
-        loading={sources.count() === 0}
+        options={getSelectOptionsFromSources(selectedDataSource)}
+        loading={selectedDataSource.count() === 0}
```
edwinmp:
```suggestion
        loading={!selectedDataSource.count()}
```

frontend/src/components/DataSourceSelector/DataSourceSelector.tsx
```diff
@@ -24,16 +35,16 @@ const DataSourceSelector: FunctionComponent<ComponentProps> = ({ source, ...prop
...
         onChange={onSelectSource}
-        value={source ? (source.get('id') as string) : undefined}
+        defaultValue={source ? (source.get('id') as string) : undefined}
```
edwinmp: Isn't this a problem? Using defaultValue & onChange? - Hope you're not getting noise in your console.

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <div className={classNames(props.className)}>
+      <label>Data Source Type</label>
+      <Dropdown
+        placeholder="Data Source Type"
+        fluid
+        selection
+        search
+        options={[
+          { value: 'non-frozen', text: 'Core Sources' },
```
edwinmp: Why name them `non-frozen`? Seems disrespectful of their coreness ... why not name the frozen ones `non-core`.

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+          { value: 'frozen', text: 'Frozen Sources' },
+        ]}
+        onChange={onSelectDataSource}
+        defaultValue={
+          props.defaultSource && props.defaultSource.length ? props.defaultSource : undefined
```
edwinmp: Default value here as well ... as long as it's not complaining.

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+        }
+        data-testid="active-data-source"
```
edwinmp: Doesn't seem right.

frontend/src/components/OperationSteps/OperationSteps.tsx
```diff
@@ -46,11 +47,16 @@ const OperationSteps: FunctionComponent<OperationStepsProps> = (props) => {
   const [isOrderingSteps, setIsOrderingSteps] = useState(false);
   const [createdSteps, setCreatedSteps] = useState<Step[]>([]);
   const sources = useSources({ limit: 200, offset: 0 });
+  const [selectedDataSource, setSelectedDataSource] = useState<List<SourceMap>>(List());
```
edwinmp: Plural.

frontend/src/components/OperationSteps/OperationSteps.tsx
```diff
@@ -146,22 +152,43 @@ const OperationSteps: FunctionComponent<OperationStepsProps> = (props) => {
     }
   };
 
+  const onDatasourceToggle = (data: string) => {
+    if (data === 'non-frozen') {
+      setSelectedDataSource(sources.filter((item) => item.get('schema') === 'repo'));
+    } else {
+      setSelectedDataSource(sources.filter((item) => item.get('schema') !== 'repo'));
+    }
```
edwinmp: Change as indicated somewhere above.

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export type SelectedDatasource = 'non-frozen' | 'frozen' | 'all';
+
+interface ComponentProps {
+  className?: string;
+  onSelect: (selectedData: SelectedDatasource) => void;
+  defaultSource?: SelectedDatasource;
+}
+
+const DataSourceSelectorToggle: FunctionComponent<ComponentProps> = ({ ...props }) => {
```
edwinmp: `DataSourceTypeSelector` ... looked down at usages and it's confusing.

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React, { FunctionComponent, useEffect } from 'react';
+import { Dropdown, DropdownProps } from 'semantic-ui-react';
+import { SelectEvent } from './DataSourceSelector';
+
+export type SelectedDatasource = 'non-frozen' | 'frozen' | 'all';
+
+interface ComponentProps {
+  className?: string;
+  onSelect: (selectedData: SelectedDatasource) => void;
+  defaultSource?: SelectedDatasource;
```
edwinmp: This is confusing too ... the value here (line 11) is not a datasource. So, `defaultDataSourceType` - even the `onSelect` is claiming to want `selectedData` ... what data? Names matter, otherwise we'd have to start adding documentation to explain stuff.

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const DataSourceSelectorToggle: FunctionComponent<ComponentProps> = ({ ...props }) => {
+  useEffect(() => {
+    if (props.defaultSource && props.defaultSource.length) {
```
edwinmp: You don't need length to determine if a string value is there or not. Strings are easily truthy or false ... so
```suggestion
    if (props.defaultSource) {
```

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+      props.onSelect(props.defaultSource);
+    }
+  }, [props.defaultSource]);
```
edwinmp: If you're watching for changes here, then this cannot be named a "default" value. So if you only care about the initial value, then this should be
```suggestion
  }, []);
```
And if that's true, then given what you're doing in here, this entire `useEffect` wouldn't be required because the default value would be available in the parent component.

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const onSelectDataSource = (_event: SelectEvent, data: DropdownProps) => {
```
edwinmp:
```suggestion
  const onSelectDataSourceType = (_event: SelectEvent, data: DropdownProps) => {
```

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
edwinmp:
```suggestion
          { value: 'core', text: 'Core Sources' },
```

frontend/src/components/OperationSteps/OperationSteps.tsx
```diff
@@ -146,22 +152,43 @@ const OperationSteps: FunctionComponent<OperationStepsProps> = (props) => {
     }
   };
 
+  const onDatasourceToggle = (data: string) => {
```
edwinmp:
```suggestion
  const onSelectDataSourceType = (dataSourceType: string) => {
```

edwinmp: You also created a type for this, so why not import it instead of using string here?

frontend/src/pages/AdvancedQueryBuilder/AdvancedQueryBuilder.tsx
```diff
@@ -30,6 +34,9 @@ const AdvancedQueryBuilder: FunctionComponent<QueryBuilderProps> = (props) => {
 
   const sources = useSources({ limit: 200, offset: 0 }) || null;
   const history = useHistory();
+  const [selectedDatasourceType, setSelectedDatasourceType] = useState<
+    'non-frozen' | 'frozen' | 'all'
```
edwinmp: Use the type definition you created.

edwinmp (changes requested):
A lot of these are opinionated since I've not actually functionally tested the feature.

edwinmp:
@davidebukali this returns some core data sources - I think perhaps the filter suggestion @wakibi is either incorrect and has some exceptions (it's mostly spotlight data sources) or it's being applied incorrectly. (see on staging where I deployed this)
image

frontend/src/components/DataSourceSelector/DataSourceSelector.tsx
davidebukali: Updated

frontend/src/components/DataSourceSelector/DataSourceSelector.tsx
davidebukali: Updated

frontend/src/components/DataSourceSelector/DataSourceSelector.tsx
davidebukali: Removed the `defaultValue` attribute

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
davidebukali: Updated

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
davidebukali: Updated

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
davidebukali: Updated

frontend/src/components/OperationSteps/OperationSteps.tsx
davidebukali: Updated

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
davidebukali: Renamed component ...

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
davidebukali: Updated

frontend/src/components/DataSourceSelector/DataSourceSelectorToggle.tsx
davidebukali: Removed unnecessary useEffect

edwinmp (approved):

PR #705, opened 2022-05-24
Fix/iati rhfp - creation of folder

dean-breed:


data_updates/Python/iati_rhfp.py
```diff
@@ -5,6 +5,7 @@
 import pandas as pd
 from github import Github, InputGitTreeElement
 
+os.mkdir("iati_csv")
```
edwinmp: This will raise a [FileExistsError](https://docs.python.org/3/library/exceptions.html#FileExistsError) error if the folder already exists so you'll need to wrap it in a [`try/except`](https://www.geeksforgeeks.org/python-os-mkdir-method/)

edwinmp (changes requested):

data_updates/Python/iati_rhfp.py
edwinmp: I've also rebased your branch so you'll need to pull before making any new changes.

dean-breed: cheers Edwin. Added now

dean-breed:
@edwinmp can we get this merged into master please?

edwinmp:
@wakibi you mentioned you'd be doing something on this ... is that still happening? If not, will proceed to deploy it.

edwinmp:
I'll assume there's nothing pending here then ... merging.

edwinmp (approved):

Issue #665, opened 2022-04-12
Progress bar


dean-breed:
When exporting something really big, all the user sees as a response is the page loading for a while, this can make them think nothing is happening - could we explore having a progress bar to make it clearer for users. This is noticeable on use of the iati transactions table

k8hughes:
Just show a loading indicator, not a count down - keep it simple

dean-breed:
This is fine by me - let me know if you have any further questions @wakibi @edwinmp

davidebukali:
@dean-breed when you get the chance, this is ready for testing on [staging](https://staging-ddw.devinit.org)

dean-breed:
@davidebukali I have taken a look at this and I am not sure the behaviour is working for large datasets. When I try to download https://staging-ddw.devinit.org/queries/data/793, it does the exporting wheel then when it gets to the save progress bar, it is an empty csv. This isn't happening on other files such as https://staging-ddw.devinit.org/queries/data/783

I am not yet clear whether this is an issue with that one query or the progress bar as a whole

davidebukali:
@dean-breed thanks for running these tests. This is a bug affecting large datasets, Ive created an issue #721; We are working on a fallback to using the old method of downloading for larger datasets until we fix this bug.

dean-breed:
Thank you for the update @davidebukali

PR #706, opened 2022-05-25
Feature | Progress bar

davidebukali:
#665

### DIFF
--- a/frontend/cypress/fixtures/exportData.json
+++ b/frontend/cypress/fixtures/exportData.json
@@ -1,0 +1,17 @@
+{
+    "errors": [],
+    "data": {
+        "country_name": "Afghanistan",
+        "di_id": "AF",
+        "iso3": "AFG",
+        "fragile_2018": "Extremely fragile"
+    },
+    "meta": {
+        "delimeter": ",",
+        "fields": ["country_name", "di_id", "iso3", "fragile_2018"],
+        "aborted": false,
+        "linebreak": "\r\n",
+        "truncated": false
+    }
+  }
+  
--- a/frontend/cypress/fixtures/exportData.json
+++ b/frontend/cypress/fixtures/exportData.json
@@ -14,4 +14,3 @@
         "truncated": false
     }
   }
-  
--- a/frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
+++ b/frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
@@ -1,8 +1,9 @@
 import { PostgreSQL, sql } from '@codemirror/lang-sql';
 import { fromJS, List } from 'immutable';
+import { parse } from 'papaparse';
 import queryString from 'query-string';
 import React, { FunctionComponent, useContext, useEffect, useState } from 'react';
-import { Button, Card, Col, Form, OverlayTrigger, Popover, Row } from 'react-bootstrap';
+import { Button, Card, Col, OverlayTrigger, Popover, Row } from 'react-bootstrap';
 import { connect, MapDispatchToProps } from 'react-redux';
 import { RouteComponentProps, useLocation, withRouter } from 'react-router-dom';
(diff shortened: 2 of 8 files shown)


frontend/cypress/fixtures/exportData.json
```diff
@@ -0,0 +1,17 @@
...
+    },
+    "meta": {
+        "delimeter": ",",
+        "fields": ["country_name", "di_id", "iso3", "fragile_2018"],
+        "aborted": false,
+        "linebreak": "\r\n",
+        "truncated": false
+    }
+  }
+
```
edwinmp: End in blank line

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
```diff
@@ -23,6 +24,9 @@ import { OperationsTableRow } from '../OperationsTableRow';
 import OperationsTableRowActions from '../OperationsTableRowActions';
 import { PaginationRow } from '../PaginationRow';
 import { SearchInput } from '../SearchInput';
+import { toast } from 'react-toastify';
+import 'react-toastify/dist/ReactToastify.css';
+import { ExportToCsv } from 'export-to-csv';
```
edwinmp: This library hasn't been updated in several years so it won't do ... let's keep things in papa parse https://stackoverflow.com/questions/52240221/download-save-csv-file-papaparse

edwinmp (changes requested):
There's another `Export to CSV` button where this feature will need to be applied ... if possible, see if you can build this in a reusable way.

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
```diff
@@ -159,6 +163,55 @@ const OperationsTableCard: FunctionComponent<OperationsTableCardProps> = (props)
...
+            fieldSeparator: ',',
+            quoteStrings: '"',
+            decimalSeparator: '.',
+            showLabels: true,
+            useTextFile: false,
+            useBom: true,
+            headers: columns && columns.length ? columns : [],
+          });
+
+          toast.update(toastId, { render: 'Success', type: 'success', isLoading: false });
```
edwinmp: If possible, account for multiple file being downloaded, so it'd be nice if the success message indicated what was successfully downloaded.

edwinmp:
@davidebukali  I deployed this to staying to see how it works & as expected from looking at the code, it doesn't show progress ... it works pretty much the same way as the old one. Only difference is that it lets you know when the download is complete by popping up an alert. This is not the expected behaviour for this feature.
As with all progress bars (progress indicator in this case), the alert is supposed to show when the process starts and goes away (or gets updated) when the process is complete.

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
davidebukali: Removed old library and now using native `js` api

frontend/src/components/OperationsTableCard/OperationsTableCard.tsx
davidebukali: Updated

frontend/cypress/fixtures/exportData.json
davidebukali: Updated

edwinmp (dismissed):
Alert Dean that this is ready for testing on staging ... the e2e tests are failing massively though... I did change the behaviour a bit so that could have broken something, but I think there's more going on overall.

edwinmp:
@davidebukali fix the conflicts on this and let's merge it in ... create a separate ticket for Dean's current bug on large downloads. Comment on the original ticket to that effect & tag the new one. Include the images I sent you in the new ticket.

davidebukali:
@edwinmp when you get the chance, this is ready for another review

edwinmp (approved):

Issue #718, opened 2022-06-14
iati_rhfp.py not writing to di-website-data repo


dean-breed:
The file designed to take IATI data from the DDW and write it to the di-website-data repo is not working. I have only seen it is not committing to the repo and not troubleshooted any further.

PR #719, opened 2022-06-14
Fix/iati rhfp bug

ThatcherK:
ticket #718

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,4 +125,3 @@ ssl/
 data_updates/manual/
 data_updates/Python/iati_csv/
 data_updates/Python/oecd_csv/
-# data_updates/Python/oecd_csv/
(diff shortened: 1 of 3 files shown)


edwinmp (commented):
@ThatcherK What exactly was the problem here and how sure are we that this has fixed it?

ThatcherK:
@edwinmp  The problem was with the local_folder argument given to the push_folder_to_github function. The csv folder was also being created in the root of the project, changed this to the Python folder.The rest were minor refactors. I am sure it works now because I tested it, and it's same as the oecd_data . But I haven't tested it with your import change

edwinmp:
Problem is that the previous version was tested as well ... anyway, will approve.

edwinmp (approved):

Issue #684, opened 2022-05-05
Add Global SDG data to the DDW


dean-breed:
The data from the global SDG database would be good to have in the DDW - https://unstats.un.org/sdgs/dataportal. I will talk to @bill-anderson when he is back from AL to discuss how best to do this. Just wanted to open up an issue while it was on the brain

bill-anderson:
See https://github.com/devinit/ddw-analyst-ui/discussions/683#discussion-4041042

dean-breed:
I have made a PR for this and adding the OECD MUMS database. I have messaged @wakibi  and made him a reviewer. Would be good to know when deployment is.

edwinmp:
Does this need to be run on dev first or is it good to get live?

dean-breed:
I am not too familiar with the process so would probably need a discussion? I have run it locally and on a local SQL server and it seems to function

PR #826, opened 2022-09-29
Feature | OECD Mums

dean-breed:
Mums and SDG for DDW next deployment. Ticket #684

### DIFF
--- a/data_updates/Python/constants.py
+++ b/data_updates/Python/constants.py
@@ -11,8 +11,10 @@ DAC1_START_YEAR=1977
 DAC1_END_YEAR=2018
 DAC2B_START_YEAR=1977
 DAC2B_END_YEAR=2018
+MUMS_START_YEAR=2011
+MUMS_END_YEAR=2018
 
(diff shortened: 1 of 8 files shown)


wakibi (changes requested):
Hi @dean-breed, can we keep all external files i.e CSVs, zips, txt files etc at this repo?
https://github.com/devinit/ddw-data-update-configs/

Also, please rename the file and remove spaces. Let's use underscores instead.

dean-breed:
@wakibi I can do both of those things but do I need to make any resulting changes due to the data being located differently?

dean-breed:
@wakibi As for the filename, I generally prefer leaving it the same as the source to make traceability more clear but if the code really needs it to not have spaces, we can do that also.

edwinmp:
I'm reopening this as I'm sure it had value - should have been check once the stale bot tagged it for closing.

edwinmp:
Guess you'll have to take this all the way @wakibi

PR #858, opened 2022-11-29
Migrate cypress

akmiller01:


akmiller01:
@edwinmp this is a mostly automatic migration to the Cypress v10 syntax. I guess dependabot upgraded cypress, but since cypress is only run on deploy, these breaking changes weren't caught.

akmiller01:
@edwinmp I've refactored this code to pass the SonarCloud filter and also fixed all passing tests. But there seems to be a genuine test failure where two of the specs attempt to delete a frozen dataset and fail with:
```
DELETE https://staging-ddw.devinit.org/api/frozendata/80/ 500 (Internal Server Error)
```

edwinmp:
> @edwinmp I've refactored this code to pass the SonarCloud filter and also fixed all passing tests. But there seems to be a genuine test failure where two of the specs attempt to delete a frozen dataset and fail with:
> 
> ```
> DELETE https://staging-ddw.devinit.org/api/frozendata/80/ 500 (Internal Server Error)
> ```

I've created a ticket for it, though it still needs looking into for the cause.

edwinmp:
@akmiller01 would love your opinion on SonarCloud before I recommend paying for it (to cover private repos).

edwinmp (approved):

akmiller01:
> @akmiller01 would love your opinion on SonarCloud before I recommend paying for it (to cover private repos).

Seems useful! A bit cumbersome to incorporate into existing projects, because touching old code triggers it to fail, but would be nice to incorporate into new projects.

Issue #899, opened 2023-02-10
Cannot enter filter value in Advanced Query Builder


akmiller01:
Steps to recreate:

1. Create new query in the Advanced Query Builder
2. Type title and description
3. Change "Select Data Source" to "International Aid Transparency Initiative Transactions"
4. Check "Filter"
5. Click "Add rule"
6. Change dropdown to "Transaction Year - Calculated"
7. Change dropdown to ">"
8. Type 2017 in the value field
9. Click "Add rule" or "Add Group"
10. Watch "2017" disappear from the value field

edwinmp:
@akmiller01 the original bug has been fixed and deployed.

@ThatcherK Caught a related bug though. Changing the column clears the existing value (even if it's still showing on the input). I'll keep the ticket open for this.


https://user-images.githubusercontent.com/5672438/219080871-a3c29dff-8f8a-46d6-bfa8-a852dcbaa006.mov

PR #904, opened 2023-02-21
Fix query and value editor mismatch on column change

ThatcherK:
Ticket #899


edwinmp:
As always, functional review first ... let me know when it's on staging.

edwinmp (approved):

Issue #688, opened 2022-05-10
Use an environment variable for a dynamic API base URL


edwinmp:
We currently have the FE and BE rigidly bound together, with the API base URL tied to the currently running instance. Using a configured value for the API base URL will give us more flexibility, allowing us to test with staging or live data without needing to deploy our code on staging or live.

PR #962, opened 2023-07-21
Use a configured base URL

APIYOJENNIFER:
Issue #688 
- Use a configured base URL to be able to test API endpoints in different environments


frontend/src/utils/api.ts
```diff
@@ -1,31 +1,33 @@
+import config from '../config';
```
ThatcherK: I don't expect any changes in this file

frontend/src/config.ts
```diff
@@ -0,0 +1,5 @@
+const config: { BASE_URL: string } = {
```
ThatcherK: This is not an environment variable..
I expect an axios config, and then the config takes in an environment variable as the baseUrl

ThatcherK (changes requested):

frontend/src/config.ts
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios';
+
+const axiosConfig = axios.create({
```
ThatcherK: This config does not seem to be referenced anywhere else apart from this file. So I doubt it's working...
Please see if [this](https://bobbyhadz.com/blog/axios-change-base-url#:~:text=%23%20Setting%20the%20axios%20Base%20URL%20by%20creating%20an%20instance,'%2C%20%7D)%3B%20placeholderApi%20.) resource helps

ThatcherK (changes requested):

ThatcherK (changes requested):
Much as this might work, I would rather you add the baseUrl in the index.js file which doesnt require you to add it to all places where axios is called. Less code. Sometrhing similar to this 
`axios.defaults.baseURL` that sets it as a global baseUrl for all axios calls. Also make sure you r environment variable is actually being read as what you have set...

frontend/src/index.ts
```diff
@@ -1 +1,4 @@
+import axios from 'axios';
 import './App';
+
+axios.defaults.baseURL = process.env.BASE_URL;
```
ThatcherK: As I sort my testing issues,move this to the App.ts file, I've seen most configs are initialised there. Let's leave the index file as clean as we found it!.. Also test that it still works. Also change the environment variable name to API_BASE_URL( its abit more descriptive)

ThatcherK (changes requested):

frontend/src/index.ts
ThatcherK: You will also need to add it to the docker_compose files. Check the environment section of the web service.
Then also add it to the readme as a variable that needs to be set

ThatcherK:
@davidebukali can you test that this works, am unable to test on my end

ThatcherK (commented):
Waiting to properly test before merging

davidebukali:
@ThatcherK this works on my side with api base url set from .env file

ThatcherK (approved):

Issue #695, opened 2022-05-17
Cypress test are failing and not always because of actual broken features


edwinmp:
Look into why this is happening for and suggest a solution & perhaps a plan for how we can better structure our e2e tests to only break for proper issues.

edwinmp:
I'm increasing the priority of this as it's falling further behind on our list of technical debt tickets.

APIYOJENNIFER:
The Cypress tests are failing intermittently due to the following issues:

- Timeout errors due to the cypress commands running before the page has fully loaded.
- Delays in loading content, which I suspect are due to large file sizes.
- Timeout errors due to slow response when making API requests.

Suggested solutions:
- Intercept all network requests and wait until it's complete before running the next cypress commands.
- Add delays where `cy.type()` is being used.

ThatcherK:
@APIYOJENNIFER Way forward on your findings

- Find all the network requests that are causing timeouts and intercept them so that all the commands that need to run after them are run after the network call is successful is made.
This resource might be helpful https://dev.to/m1073496/cypress-and-flaky-tests-how-to-handle-timeout-errors-4ggm

ThatcherK:
All tests have been fixed apart from the `deletes a frozen datasource` test under datasources, where fixing the timeout errors exposed an api error that is referenced in this issue #970

PR #967, opened 2023-08-09
Fix | Intermittent cypress tests failures

APIYOJENNIFER:
Issue #695 
**Datasets**
- Add wait when typing in the search bar to allow the list items to get loaded.
- Delete a frozen dataset by using two `click()` commands.

**QueryBuilder.select**
- Add delay to the `cy.type()` command in the _commands.js_ file to fix timeout errors in _queryBuilder.select_ file.

**Datasources**
- Navigate to the sources URL before attempting to get the `.dataset-row` to fix the timeout error.

### DIFF
--- a/frontend/cypress/integration/dataSources.cy.js
+++ b/frontend/cypress/integration/dataSources.cy.js
@@ -124,26 +124,35 @@ describe('The Data Sources Page', () => {
     });
 
     it('downloads a successfully frozen data source', () => {
-      cy.wait(100);
-      cy.get('.dataset-row')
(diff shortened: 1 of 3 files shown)


wakibi (approved):
LGTM

ThatcherK (approved):
@APIYOJENNIFER good job on this!! And for exposing the bug on the `deletes a frozen data source` endpoint. I will add that as a note on the issue. So the issue will stay open , so that we don't forget about it

ThatcherK:
Am actually going to close the issue, the remaining test case will be handled in this issue #970

Issue #703, opened 2022-05-23
Hide/Remove versions action for frozen data sources


edwinmp:
<img width="1220" alt="image" src="https://user-images.githubusercontent.com/5672438/169755934-0e3f9d2e-6ec9-4a54-b37b-68508edde12a.png">

**Requirements**

- [ ] On the data sources page, table rows for `Frozen Sources` should not have the `Versions` button for any filter that renders frozen sources i.e `Frozen Sources` and `All Sources`
- [ ] Update e2e tests to include one for testing this feature & for the filter itself (wasn't done on its implementation)

**Completion Criteria**

- [ ] `Versions` button is not available for frozen sources
- [ ] e2e tests (new & old) are running successfully

edwinmp:
Not high priority but for consistency's sake, we should still do this sometime soon.

PR #978, opened 2023-08-25
Fix | Hide/Remove Versions Button, Tests

APIYOJENNIFER:
Fixes issue #703 
- Hide the Version button for Frozen and All sources
- End-to-end test for asserting that the Versions button is hidden
- End-to-end test for validating the dropdown filter

### DIFF
--- a/frontend/src/components/SourcesTableCard/SourcesTableCard.tsx
+++ b/frontend/src/components/SourcesTableCard/SourcesTableCard.tsx
@@ -21,7 +21,7 @@ interface ComponentProps {
   offset: number;
   links?: LinksMap;
   count: number;
-  value: number;
+  value?: number;
 }
 type SourcesTableCardProps = ComponentProps;
 type TableFilters = {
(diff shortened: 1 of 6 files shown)


frontend/src/components/SourcesTableCard/SourcesTableCard.tsx
```diff
@@ -131,12 +132,17 @@ export const SourcesTableCard: FunctionComponent<SourcesTableCardProps> = (props
...
               />
             </Col>
           </Row>
         </Card.Body>
         <Card.Body>
-          <SourcesTable sources={props.sources} activeSource={props.activeSource} />
+          <SourcesTable
+            sources={props.sources}
+            activeSource={props.activeSource}
+            value={frozenQuery}
```
ThatcherK: Could we give this prop a different name, value kind of suggests an input perhaps name it `sourceFilterValue` just for more context

ThatcherK (changes requested):

ThatcherK (approved):

PR #1053, opened 2024-02-28
Pin Pandas version and move back to native to_sql

akmiller01:


data_updates/Python/iati_transactions.py
```diff
@@ -201,16 +193,8 @@ def main(args):
...
-        flat_transaction_data_batches_generator = batch_generator(flat_transaction_data_records)
         with engine.begin() as conn:
-            for flat_transaction_data_batch in flat_transaction_data_batches_generator:
-                conn.execute(
-                    insert(tmp_transaction_table).values(
-                        flat_transaction_data_batch
-                    )
-                )
-
+            flat_transaction_data.to_sql(name=TMP_DATA_TABLENAME, con=conn, schema=DATA_SCHEMA, index=False, if_exists="append")
```
wakibi, line 197: Won't this create a very long sql statement?

akmiller01: > Won't this create a very long sql statement?

It was the way all our Python scripts worked before SQL alchemy got updated from v1 to v2. Pandas v1 was incompatible with SQL alchemy v2. The batching function I wrote to replace pandas `to_sql` ended up taking about 5x as long to run as the original. To the point where large updates would take ~30 hours.

So we need to update Python from 3.7 to 3.8, in order to pull in pandas v2 and get the original fast `to_sql` back. It does smart batching.

wakibi: If it does batching, we good.

wakibi (approved):

